PR #17, opened 2016-11-14
Accounts login

TimMikeladze:
This PR contains the core code for the `@accounts/accounts`, Accounts options, error handling, implementation of the login function, and a naive logout function.


src/Accounts.js
```diff
@@ -0,0 +1,517 @@
+/* eslint-disable max-len */
+
+import { each, keys, includes, trim, isEmpty, merge, extend } from 'lodash';
+import { Iterable } from 'immutable';
```
alexyuly: You could use Immutable throughout your redux implementation. Some of the reducer cases (below) could be written more tersely. Are you not doing so, in order to keep this package more generic for consumers not using immutable?

src/Accounts.js
```diff
@@ -0,0 +1,517 @@
...
+      break;
+    }
+    case CLEAR_FORM: {
+      const { form } = action.payload;
+      nextState = extend({}, state, {
+        forms: {
+          [form]: initialState.forms[form],
+        },
+      });
+      break;
```
alexyuly: Style nitpick - I prefer to avoid `let` and `break` and return directly from each case, but that may be a style choice on your part in this package.

src/Accounts.js
```diff
@@ -0,0 +1,517 @@
...
+        throw new Error(`Accounts.ui.config: Can't set \`requestOfflineToken\` more than once for ${service}`);
+      } else {
+        Accounts.ui._options.requestOfflineToken[service] = value;
+      }
+    });
+  }
+
+  // deal with `forceApprovalPrompt`
+  if (options.forceApprovalPrompt) {
+    each(options.forceApprovalPrompt, (value, service) => {
```
alexyuly: Why use this over native `forEach`?

src/Accounts.js
TimMikeladze: Yeah I could and would prefer to. Not using immutable was an effort to stay vanilla because I wasn't sure of what consumption of this would be like. It's possible to use an immutable store with Accounts already, I'll make the additional changes to use immutable.

PR #24, opened 2016-12-18
docs(Design): Started a design document. Added packages structure

davidyaha:
Hey,

Added a simple diagram that shows the package structure as I understand it.
This PR is meant to hold feedback on the project structure.

Some things are left out.. For example the packages relations (what package depends on which package) is left out and should be discussed as well.

As I see it, the dependency tree should look like so:
Client: View -> Client Logic -> Transport
Server: Server Logic -> Transport, Data store

Would love some feedback here.


TimMikeladze:
@davidyaha Your understanding of the package structure is correct. 

1. An addition I would make to the server side would be the http servers (express, koa, hapi). We need to provide the necessary middleware to setup the routes or graphql depending on the user's choice of framework. I'm not sure whether these should reside in `@accounts/accounts`, in the server side transports, or as their own packages. 

2.  The diagram is missing the oauth portion. https://github.com/simov/grant is the package I was intending on using. This also means that we need to provide callback routes for the oauth flow as part of the middleware.

davidyaha:
1. Yeah so I was thinking about it earlier and I think I missed crucial layer which is the actual network transport/protocol to use with REST or GraphQL.
Theoretically speaking, both can use HTTP or WebSockets or any other cloud based transport like [Pusher](http://pusher.com) or any other thing that someone will come up with.
So basically I propose having something like this on the diagram
------------------------------------------
Data fetching Client (GraphQL, Rest, DDP)
------------------------------------------
Transport (HTTP (express, koa, etc..)/WS)
------------------------------------------
Data fetching API (GraphQL, Rest, DDP)
------------------------------------------

2. I was thinking oauth would be implemented as part of the client/server logic that would basically  be every package used and exposed by the accounts package. Your thoughts?

TimMikeladze:
1. Looks good.
2. Could you please expound on that? The way I see it is `@accounts/accounts` provides the logic to for oauth logins (this includes initiating the request and creating/logging in as the user via the provided db package). Each of the server side transports (express, koa, hapi) need to provide callback routes for the oauth flow, this is achieved by the middleware provided from the `grant` package.

davidyaha:
@TimMikeladze Update the diagram.
I think we should merge that and the current devel branch and start working on an example. What do you think?

TimMikeladze:
@davidyaha Diagram looks good. I have some additional work to do on devel today and then I'll send a PR for review.

PR #25, opened 2016-12-20
Server side Accounts.createUser

TimMikeladze:
This PR introduces the Accounts.createUser function, tests, and beginnings of the DBDriverInterface, and a variety of minor fixes.

### DIFF
--- a/src/client/AccountsClient.js
+++ b/src/client/AccountsClient.js
@@ -10,13 +10,13 @@ import AccountsCommon from '../common/AccountsCommon';
 import { loggingIn } from './module';
 import ui from './ui';
 import type { AccountsOptionsType } from '../common/AccountsCommon';
+import type UserObjectType from '../common/UserObjectType';
 
-export type UserObjectType = {username: ?string, email: ?string, id: ?string};
 const isValidUserObject = (user: UserObjectType) => has(user, 'user') || has(user, 'email') || has(user, 'id');
 
 export type UserCreationInputType = {
   username: ?string,
-  password: ?string,
+  password: string,
   email: ?string,
   profile: ?Object
 };
@@ -50,11 +50,11 @@ class Accounts extends AccountsCommon {
   }
 
   // TODO Accept 'profile' in the options
-  async createUser(user: UserCreationInputType, callback: Function): Promise<void> {
+  async createUser(user: UserCreationInputType, callback: ?Function): Promise<void> {
     this.validatePassword(user.password);
     // TODO Throw error if client user creation is disabled
 
-    if (!this.validateUsername(user.username, false) && !this.validateEmail(user.email, false)) {
+    if (this.validateUsername(user.username, false) && this.validateEmail(user.email, false)) {
       throw new AccountsError({ message: 'Username or Email is required' });
     }
     try {
--- a/src/client/AccountsClient.js
+++ b/src/client/AccountsClient.js
@@ -54,7 +54,7 @@ class Accounts extends AccountsCommon {
     this.validatePassword(user.password);
     // TODO Throw error if client user creation is disabled
 
-    if (this.validateUsername(user.username, false) && this.validateEmail(user.email, false)) {
+    if (!this.validateUsername(user.username, false) && !this.validateEmail(user.email, false)) {
       throw new AccountsError({ message: 'Username or Email is required' });
     }
     try {
--- a/src/common/UserObjectType.js
+++ b/src/common/UserObjectType.js
@@ -1,0 +1,6 @@
+export type UserObjectType = {
+  username: ?string,
+  email: ?string,
+  id: ?string,
+  profile: ?Object,
+};
--- a/src/server/AccountsServer.js
+++ b/src/server/AccountsServer.js
@@ -28,7 +28,7 @@ class Accounts extends AccountsCommon {
     this.db = db;
   }
   async createUser(user: UserCreationInputType): Promise<string> {
-    if (this.validateUsername(user.username, false) && this.validateEmail(user.email, false)) {
+    if (!this.validateUsername(user.username, false) && !this.validateEmail(user.email, false)) {
       throw new AccountsError({ message: 'Username or Email is required' });
     }
     if (user.username && await this.db.findUserByUsername(user.username)) {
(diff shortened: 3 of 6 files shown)


src/client/AccountsClient.js
```diff
@@ -10,13 +10,13 @@ import AccountsCommon from '../common/AccountsCommon';
...
 import type { AccountsOptionsType } from '../common/AccountsCommon';
+import type UserObjectType from '../common/UserObjectType';
 
-export type UserObjectType = {username: ?string, email: ?string, id: ?string};
 const isValidUserObject = (user: UserObjectType) => has(user, 'user') || has(user, 'email') || has(user, 'id');
 
 export type UserCreationInputType = {
   username: ?string,
-  password: ?string,
+  password: string,
```
davidyaha: I think this is supposed to be nullable because of oauth or strategies where password is not required like phone number validation (Send a text to your phone number and realise this device is to be authenticated)
Another option would be to have these kind of strategies handled by different methods.

LMK what you think.

src/client/AccountsClient.js
```diff
@@ -50,11 +50,11 @@ class Accounts extends AccountsCommon {
   }
 
   // TODO Accept 'profile' in the options
-  async createUser(user: UserCreationInputType, callback: Function): Promise<void> {
+  async createUser(user: UserCreationInputType, callback: ?Function): Promise<void> {
     this.validatePassword(user.password);
     // TODO Throw error if client user creation is disabled
 
-    if (!this.validateUsername(user.username, false) && !this.validateEmail(user.email, false)) {
+    if (this.validateUsername(user.username, false) && this.validateEmail(user.email, false)) {
```
davidyaha: The validateXXX methods should return true whenever email is valid and not when there's an error in my opinion. If change that, the logic you had before will be correct.

src/common/UserObjectType.js
```diff
@@ -0,0 +1,6 @@
+export type UserObjectType = {
+  username: ?string,
+  email: ?string,
+  id: ?string,
+  profile: ?Object,
```
davidyaha: Consider adding generics to this type to give the user an option to have an extensible UserObject.

src/server/AccountsServer.js
```diff
@@ -4,9 +4,21 @@ import { defaultServerConfig } from '../common/defaultConfigs';
 import { AccountsError } from '../common/errors';
 import AccountsCommon from '../common/AccountsCommon';
 import type { AccountsOptionsType } from '../common/AccountsCommon';
+import type { DBDriverInterface } from './DBDriver';
+import type UserObjectType from '../common/UserObjectType';
+
+
+export type UserCreationInputType = {
```
davidyaha: Maybe move that into the common folder so both could the client and the server will use the same type?

Also a possibility will be to create declaration files.

src/server/AccountsServer.js
```diff
@@ -15,17 +27,38 @@ class Accounts extends AccountsCommon {
     }
     this.db = db;
   }
+  async createUser(user: UserCreationInputType): Promise<string> {
+    if (this.validateUsername(user.username, false) && this.validateEmail(user.email, false)) {
+      throw new AccountsError({ message: 'Username or Email is required' });
+    }
+    if (user.username && await this.db.findUserByUsername(user.username)) {
```
davidyaha: I think a better approach here would be to let the `dbInterface` have a general `findUser` method that will accept `username`, `email`, `id`, `loginToken` or whatever identifier. This way the logic for how the data store holds his users and finds them is not tied to this `createUser` logic. 

Some `dbInterfaces` might be able to do this check with only one round trip to the actual db.

src/server/AccountsServer.spec.js
```diff
@@ -23,4 +26,72 @@ describe('Accounts', () => {
...
+    it('throws error if username exists', async () => {
+      Accounts.config({}, {
+        ...db,
+        findUserByUsername: () => Promise.resolve('user'),
+      });
+      try {
+        await Accounts.createUser({
+          password: '123456',
+          username: 'user1',
+          email: '',
```
davidyaha: Maybe add a test case where email or username is undefined? This should still work as long as one of them is a valid filed.

src/server/DBDriver.js
```diff
@@ -1,3 +1,12 @@
+// @flow
+import type UserObjectType from '../common/UserObjectType';
+
+export interface DBDriverInterface {
+  findUserByEmail(email: string): Promise<UserObjectType> | Promise<null>,
```
davidyaha: As talked on slack, use `Promise<?UserObjectType>` for nullables

src/client/AccountsClient.js
TimMikeladze: I'm following the Meteor accounts documentation. On the client calling Accounts.createUser a password is required. On the server Accounts.createUser can be called without a password but the user wont be able to login until Accounts.setPassword is called. 

Accounts.createUser isn't used for oauth or other strategies, only for passwords. The other strategies will be handled by another method, probably Accounts.createUserFromService

src/common/UserObjectType.js
TimMikeladze: Hmmm. I'm not sure. What would the usage scenario of this be? If the consuming code wants to extend the user object is should be handled through Accounts.onCreateUser http://docs.meteor.com/api/accounts-multi.html#AccountsServer-onCreateUser

src/server/AccountsServer.js
TimMikeladze: Passwords are optional on the server but they are not optional on the client so we need to have two different input types.

src/server/AccountsServer.js
TimMikeladze: I agree. I was following Meteor's documentation which has the findUserByUsername/Email methods. However your suggested approach makes more sense.

src/server/AccountsServer.spec.js
TimMikeladze: Good idea.

src/server/DBDriver.js
TimMikeladze: Yes!

src/client/AccountsClient.js
TimMikeladze: Good call. I'm dumb.

src/server/AccountsServer.js
TimMikeladze: Actually, we lose granularity of error messages if it's done in this fashion.

src/common/UserObjectType.js
davidyaha: This type should describe the returned object from any method that should give you a user object. Meaning IMO that the profile field could be defined by the using application however it would like. This is true for meteor as well. Generics will allow the user to have auto completion for his own defined profile object.. Not a must, but could be cool to have.

davidyaha:
Good to merge IMO.

PR #28, opened 2016-12-26
Setup jest

pradel:
Setup jest


src/server/AccountsServer.spec.js
```diff
@@ -79,7 +75,7 @@ describe('Accounts', () => {
         expect.fail();
```
TimMikeladze: Is `expect.fail()` part of the jest api? If not we need to either a) trigger a failure or b) rewrite these test cases to use https://facebook.github.io/jest/docs/api.html#tothrowerrorerror.

PR #34, opened 2017-01-16
Token and session management

TimMikeladze:


src/client/AccountsClient.js
```diff
@@ -48,13 +53,75 @@ export class AccountsClient {
...
+    localStorage.removeItem(REFRESH_TOKEN);
+  }
+  clearUser() {
+    this.store.dispatch(clearUser());
+  }
+  async resumeSession(): Promise<void> {
+    const { accessToken } = this.tokens();
+    if (accessToken) {
+      try {
+        const decodedAccessToken = jwtDecode(accessToken);
```
BradfordMedeiros: Is this the right way to do this?   We're technically validating the login here on the client, correct?   This means with a malicious client we can never expire them since we only check the expiration via the client's token.

This is my first time looking at this code, so definitely could be overlooking something.

src/client/AccountsClient.js
```diff
@@ -48,13 +53,75 @@ export class AccountsClient {
...
+        const currentTime = Date.now() / 1000;
+        if (decodedAccessToken.exp < currentTime) {
+          // Access token is expired, try to request a new token pair
+          await this.refreshSession();
+        } else { // Access token is still valid, resume the session
+          this.store.dispatch(setUser(decodedAccessToken.data.user));
```
BradfordMedeiros: Are we storing all the data locally here?

src/client/AccountsClient.js
TimMikeladze: The access token expires by itself over time. The access and refresh tokens are created and stored server side.

src/client/AccountsClient.js
TimMikeladze: When the access token is created the essential user data "id, username, email" are encoded within it and returned to the client. This data is also stored server side.

BradfordMedeiros: where do we actually send the data to the server?

TimMikeladze: The `this.transport.someMethod` calls send data to the server

PR #35, opened 2017-01-17
Add configurable prefix to avoid local storage key collisions on the same domain

TimMikeladze:


TimMikeladze:
That's right. I was thinking it was gonna be harder to solve then it turned out to be, but it was a real "wtf" moment at first.

PR #36, opened 2017-01-18
Logout and session improvements

TimMikeladze:
This PR introduces code for server side logout. It also improves the session code by only encoding the session id within access tokens to avoid leaking any user information. Also the resume session code is now handled by the refreshTokens method since these are essentially the same thing.


TimMikeladze:
These changes will come in as part of the Lerna PR.

PR #39, opened 2017-01-24
Unify lint config

davidyaha:
Moved eslint config file to the root directory and changed it's script so it will run with that config on all packages.


TimMikeladze:
@davidyaha This looks good. Can you give the other lerna repos a similar treatment please?

PR #41, opened 2017-01-26
Refactor errors

### LINKED ISSUE #31: Rewrite / Improve error handling
Currently the `apollo-errors` package is being consumed by our custom `AccountsError` class to throw errors. Moving forward we can continue using `apollo-errors` or make our own custom error class. The reason behind using `apollo-errors` is that it allows errors to carry more data than just an error message, this is particularly useful in the case of validation errors where you want to associate errors with certain fields. Furthermore this allows http status codes to be sent along side the message, rather than within it. 

The other packages that consume `@accounts/accounts`: `rest-api`, `rest`, `react` and the db interfaces will also need their error handling code to be updated as part of this issue.

davidyaha:
I removed our need for apollo-errors. It actually does not make sense to use that package on this non apollo package.
If this error class seems good to you I will add implement the changes needed to use for our other packages.

Error code in my opinion should not be HTTP codes as they are transport specific really.
I suggest we will add an error dictionary that will then be extended with usable ui strings on the view layer packages.

login info is passed where available in order to allow audit mechanism of some sort. This is low priority feature but I will add that to our other issues so we will remember that for the future (If it does not already exists over there) 

Closes #31


packages/common/src/errors.spec.js
```diff
@@ -0,0 +1,51 @@
...
+    } catch (e) {
+      expect(e.message).toBe('Validation Error');
+      expect(e.loginInfo).toMatchObject({ username: 'user' });
+      expect(e.errorCode).toBe('ACCOUNTS:1');
+      expect(e.epochTime).toBeDefined();
+    }
+  });
+
+
+  // NB: For some reason it seems jest have different Error class
```
TimMikeladze: Maybe an approach leveraging http://facebook.github.io/jest/docs/api.html#extending-jest-matchers could be possible to produce something alone the lines of `expect(function).toThrowAccountsError()`.

packages/common/src/errors.js
```diff
@@ -1,6 +1,41 @@
...
+import type { PasswordLoginUserType } from './types';
 
 // eslint-disable-next-line import/prefer-default-export
-export { AccountsError };
+export class AccountsError extends Error {
+  loginInfo: PasswordLoginUserType;
+  errorCode: string | number;
+  epochTime: number;
+
+  constructor(message: string, loginInfo?: PasswordLoginUserType, errorCode?: string | number) {
```
TimMikeladze: The second parameter, `loginInfo` should be more generally named and typed, other data is passed into here too.

TimMikeladze:
Great call on adding an error dictionary. Could you make an issue for that?

packages/common/src/errors.js
davidyaha: Actually I would like this parameter to be only the login info for later use in the audit system. We can add in additionalData field for other things as well, what kind of data are you thinking of sending? 
Also, keep in mind that you always use string templates to add in relevant data to the error message. `message` filed will actually get picked up by any console or error handler like react-native's "red screen" and should actually help with fixing bugs when using this package.

packages/common/src/errors.spec.js
davidyaha: I will look into that! Thanks!

davidyaha:
Sure NP

packages/common/src/errors.js
TimMikeladze: Alright.

davidyaha: 👍

davidyaha:
@TimMikeladze Fixed the issue with jest or rather with babel not properly extending `Error` class. Merge?

PR #45, opened 2017-01-30
Sessions and user info

TimMikeladze:
This PR introduces a variety of improvements to the user session code and also adds support for user profiles.

### DIFF
--- a/packages/server/src/AccountsServer.js
+++ b/packages/server/src/AccountsServer.js
@@ -302,6 +302,9 @@ const Accounts = {
   resumeSession(accessToken: string): Promise<UserObjectType> {
     return this.instance.resumeSession(accessToken);
   },
+  userId(): string | null {
+    return null;
+  },
 };
 
 export default Accounts;
--- a/packages/server/src/AccountsServer.js
+++ b/packages/server/src/AccountsServer.js
@@ -174,6 +174,36 @@ export class AccountsServer {
     return { accessToken, refreshToken };
   }
   async logout(accessToken: string): Promise<void> {
+    const session : SessionType = await this.findSessionByAccessToken(accessToken);
+
+    if (session.valid) {
+      const user = await this.db.findUserById(session.userId);
+      if (!user) {
+        throw new AccountsError({
+          message: 'User not found',
+        });
+      }
+      await this.db.invalidateSession(session.sessionId);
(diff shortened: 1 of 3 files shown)


packages/server/src/AccountsServer.js
```diff
@@ -174,6 +174,35 @@ export class AccountsServer {
...
+      }
+      await this.db.invalidateSession(session.sessionId);
+    } else { // eslint-disable-line no-else-return
+      throw new AccountsError({
+        message: 'Session is no longer valid',
+      });
+    }
+  }
+  async resumeSession(accessToken: string): Promise<UserObjectType> {
+    const session : UserObjectType = await this.findSessionByAccessToken(accessToken);
```
pradel: SessionType instead of SessionType

packages/server/src/AccountsServer.js
```diff
@@ -174,6 +174,35 @@ export class AccountsServer {
...
+        });
+      }
+      await this.db.invalidateSession(session.sessionId);
+    } else { // eslint-disable-line no-else-return
+      throw new AccountsError({
+        message: 'Session is no longer valid',
+      });
+    }
+  }
+  async resumeSession(accessToken: string): Promise<UserObjectType> {
```
pradel: ?UserObjectType could return null

packages/server/src/AccountsServer.js
```diff
@@ -174,6 +174,35 @@ export class AccountsServer {
...
+    const session : SessionType = await this.findSessionByAccessToken(accessToken);
+    if (session.valid) {
+      const user = await this.db.findUserById(session.userId);
+      if (!user) {
+        throw new AccountsError({
+          message: 'User not found',
+        });
+      }
+      await this.db.invalidateSession(session.sessionId);
+    } else { // eslint-disable-line no-else-return
```
pradel: Should we delete the invalid session if it's not longer valid ?

TimMikeladze: I think the lifespan of expired sessions should be configurable. It can provide a useful log for user account activity. Keep the latest 10 sessions for example.

TimMikeladze: I'll create an issue for this.

TimMikeladze:
@pradel Updated.

TimMikeladze:
Thanks!

PR #47, opened 2017-02-01
Verify email

pradel:


packages/server/src/DBInterface.js
```diff
@@ -12,6 +12,7 @@ export interface DBInterface {
   findUserByEmail(email: string) : Promise<?UserObjectType>,
   findUserByUsername(username: string) : Promise<?UserObjectType>,
   findUserById(userId: string) : Promise<?UserObjectType>,
+  findUserByEmailVerificationToken(token: string) : Promise<?UserObjectType>,
```
pradel: Not sure about the name of the function

packages/server/src/tokens.js
```diff
@@ -1,5 +1,8 @@
 // @flow
 import jwt from 'jsonwebtoken';
+import crypto from 'crypto';
+
+export const generateToken = (length: Int = 43) => crypto.randomBytes(length).toString('hex');
```
pradel: Is it a good place for this function?

packages/server/src/AccountsServer.js
```diff
@@ -242,8 +242,27 @@ export class AccountsServer {
...
-  verifyEmail(userId: string, email: string): Promise<void> {
-    return this.db.verifyEmail(userId, email);
+  async verifyEmail(token: string): Promise<void> {
+    const user = await this.db.findUserByEmailVerificationToken();
+    if (!user) {
+      throw new AccountsError({
+        message: 'Verify email link expired',
+      });
+    }
+    const tokenRecord = find(user.services.email.verificationTokens,
```
pradel: @TimMikeladze will all dbInterfaces return the user object formatted in the same way ?

TimMikeladze: Yes they should.

packages/server/src/tokens.js
TimMikeladze: Good place. I'd call it generateEmailToken.

packages/server/src/AccountsServer.js
```diff
@@ -242,8 +242,27 @@ export class AccountsServer {
...
+                             (t: Object) => t.token === token);
+    if (!tokenRecord) {
+      throw new AccountsError({
+        message: 'Verify email link expired',
+      });
+    }
+    const emailRecord = find(user.emails, (e: Object) => e.address === tokenRecord.address);
```
pradel: I think it will be the only place we will need this function, I don't think it's necessary to make a special function.

PR #50, opened 2017-02-05
feat(authenticator): add custom password authenticator

dotansimha:
Added support for custom password authenticator. 
Useful is you are using `accounts` to manage your users, but using an external authenticator module or server (for example - LDAP / AD).
You can still manage the creation of users and it's management with `accounts`, but the password is done using an external service. 

Example for ActiveDirectory (using `activedirectory` from npm):
```js
const ad = new ActiveDirectory(config);

AccountsServer.config({
  passwordAuthenticator: async (user, password) => {
    return new Promise((resolve, reject) => {
      const fail = () => {
        reject('Invalid username/password');
      };

     const user = user.username || user.email;

      ad.authenticate(user, password, function(err, auth) {
        if (err) {
          return fail();
        }

        if (auth) {
          ad.findUser(user, (err, userData) => {
            if (err) {
              return fail();
            }

           // Here you can also manage the creation of the user, the goal is to resolve and return the user's object (`foundUser`) from here, then `accounts` does the rest and manages the token!

            resolve(userData);
          });
        }
        else {
          return fail();
        }
      });
    });
  }
}, new RedisDBInterface());

```


packages/server/src/AccountsServer.js
```diff
@@ -50,41 +50,71 @@ export class AccountsServer {
...
+      user: foundUser,
+      tokens: {
+        refreshToken,
+        accessToken,
+      },
+    };
+  }
+
+  async _externalPasswordAuthenticator(authFn, id, username, email, password) {
+    return await authFn(id, username, email, password, password);
```
TimMikeladze: `password` is being twice in pass.

Consider changing the function arguments to `user: PasswordLoginUserType, password: string` similar to `AccountsServer.loginWithPassword`

packages/server/src/AccountsServer.js
```diff
@@ -50,41 +50,71 @@ export class AccountsServer {
       : toUsernameAndEmail({ ...user });
 
     let foundUser;
+
+    if (this.options.passwordAuthenticator) {
```
TimMikeladze: Please add / modify relevant test cases.

packages/server/src/AccountsServer.js
dotansimha: Sure! done

packages/server/src/AccountsServer.js
dotansimha: Done!

TimMikeladze:
Good merge?

PR #53, opened 2017-02-06
bugfix(id): fix a bug with session created without userId when using …

dotansimha:
…external authenticator


davidyaha:
Talked about it offline and we should transform the document on the db layer to have `id` field rather than expect it to be `_id` or `id`

pradel:
It is now fixed on latest mongo release

PR #49, opened 2017-02-05
Return directly a new AccountsServer object

pradel:
Remove the interface

PR #57, opened 2017-02-09
Config need to throw when no db object is passed

pradel:
It look like I remove this one on #49 sorry!
I updated the test, now it will fail if the check is missing.


TimMikeladze:
Ah nice catch.

Issue #40, opened 2017-01-25
Emails and password reset


TimMikeladze:
The following should be implemented, additional details can be found here http://docs.meteor.com/api/passwords.html. 

An important detail to note in the email implementation is that should be generic enough for consumers to plugin their own email service. For example using the mailgun api vs using their own smtp server.

**Server**
- [x] addEmail
- [x] removeEmail
- [x] sendResetPasswordEmail
- [ ] sendEnrollmentEmail
- [x] sendVerificationEmail
- [x] resetPassword
- [ ] sendWelcomeEmail

**Server config**
- [ ] sendVerificationEmail (default false)
- [ ] sendEnrollmentEmail (default false)
- [ ] sendWelcomeEmail (default false)
- [ ] email template for each email type, consider using https://github.com/mjmlio/mjml

**Client**
- [ ] requestPasswordReset
- [ ] requestVerificationEmail

**Client config**
- [ ] onResetPasswordLink
- [ ] onEnrollmentLink
- [ ] onEmailVerificationLink

pradel:
For the mail implementation I found this lib https://github.com/eleith/emailjs who looks good. I was using [nodemailer](https://github.com/nodemailer/nodemailer) but they changed their licence since v3

pradel:
Btw I am not sure we should provide email templates.
I think we could have only text emails and let the user do his own template.

TimMikeladze:
Sounds good to me.

PR #59, opened 2017-02-15
sendVerificationEmail and setup email

pradel:
#40 
Setup email part.
Add siteUrl in config to construct absolute urls for email.
Add sendVerificationEmail.
Add sendResetPasswordEmail.

To overwrite the email templates:
```javascript
AccountsServer.config({
 siteUrl: 'https://my-app.com',
 email: // a valid email config object passed to emailjs
 // https://github.com/eleith/emailjs#example-usage---text-only-emails
});

AccountsServer.emailTemplates.from = 'my-app <no-reply@my-app.com>';
AccountsServer.emailTemplates.verifyEmail.subject = (user) => `Verify your account email ${user.profile.lastname}`;
// Same for AccountsServer.emailTemplates.verifyEmail.text = (user, url)
```

### DIFF
--- a/packages/server/src/AccountsServer.js
+++ b/packages/server/src/AccountsServer.js
@@ -21,6 +21,7 @@ import { verifyPassword } from './encryption';
 import {
   generateAccessToken,
   generateRefreshToken,
+  generateEmailToken,
 } from './tokens';
 import Email from './email';
 import emailTemplates from './emailTemplates';
@@ -335,14 +336,15 @@ export class AccountsServer {
     if (!address || !includes(user.emails.map((email: string) => email.address), address)) {
       throw new AccountsError('No such email address for user');
     }
-    const token = await this.db.addEmailVerificationToken(userId, address);
-    const resetPasswordUrl = `${this._options.siteUrl}/verify-email/${token}`;
+    const token = generateEmailToken();
+    await this.db.addEmailVerificationToken(userId, address, token);
+    const verifyEmailUrl = `${this._options.siteUrl}/verify-email/${token}`;
     await this.email.sendMail({
       from: this.emailTemplates.verifyEmail.from ?
         this.emailTemplates.verifyEmail.from : this.emailTemplates.from,
       to: address,
       subject: this.emailTemplates.verifyEmail.subject(user),
-      text: this.emailTemplates.verifyEmail.text(user, resetPasswordUrl),
+      text: this.emailTemplates.verifyEmail.text(user, verifyEmailUrl),
     });
   }
 }
--- a/packages/server/src/AccountsServer.js
+++ b/packages/server/src/AccountsServer.js
@@ -39,7 +39,11 @@ export class AccountsServer {
       throw new AccountsError('A database driver is required');
     }
     this.db = db;
-    this.email = new Email(config.email);
+    if (options.sendMail) {
+      this.email = { sendMail: options.sendMail };
+    } else {
+      this.email = new Email(config.email);
+    }
     this.emailTemplates = emailTemplates;
   }
 
--- a/packages/server/src/AccountsServer.js
+++ b/packages/server/src/AccountsServer.js
@@ -383,7 +383,7 @@ export class AccountsServer {
       throw new AccountsError('No such email address for user');
     }
     const token = generateRandomToken();
-    await this.db.addPasswordResetToken(userId, address, token);
+    await this.db.addResetPasswordToken(userId, address, token);
     const resetPasswordUrl = `${this._options.siteUrl}/reset-password/${token}`;
     await this.email.sendMail({
       from: this.emailTemplates.resetPassword.from ?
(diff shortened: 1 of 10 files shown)


packages/server/src/AccountsServer.js
```diff
@@ -302,6 +320,31 @@ export class AccountsServer {
...
+    // If no address provided find the first unverified email
+    if (!address) {
+      const email = find(user.emails, (e: string) => !e.verified);
+      address = email && email.address; // eslint-disable-line no-param-reassign
+    }
+    // Make sure the address is valid
+    if (!address || !includes(user.emails.map((email: string) => email.address), address)) {
+      throw new AccountsError('No such email address for user');
+    }
+    const token = await this.db.addEmailVerificationToken(userId, address);
```
davidyaha: The db should layer should probably not be responsible of creating the token as this will be the same logic for every store. The only caveat with moving that logic to here is that the token should be unique and so we need the db to allow search by email verification token to make sure the "new token" is valid.

packages/server/src/AccountsServer.js
```diff
@@ -302,6 +320,31 @@ export class AccountsServer {
...
+    const resetPasswordUrl = `${this._options.siteUrl}/reset-password/${token}`;
```
davidyaha: Why this has reset-password route?

packages/server/src/AccountsServer.js
pradel: Right now there is this function that generate the token https://github.com/js-accounts/accounts/blob/master/packages/server/src/tokens.js#L5 and in verifyEmail function there is this call to the bdd (who must take the token as parameter btw) to find the user with the token

pradel:
@davidyaha Now the token is generated in the function, is it better?

packages/server/src/AccountsServer.js
```diff
@@ -302,6 +321,57 @@ export class AccountsServer {
     const res = await this.db.setProfile(userId, { ...user.profile, ...profile });
     return res;
   }
+
+  async sendVerificationEmail(userId: string, address: string): Promise<void> {
+    const user = await this.db.findUserById(userId);
```
davidyaha: You should probably extract the code from this line to L338 to another function because it is repeating on the `sendResetPasswordEmail` method

pradel: It is not exactly the same things for `sendVerificationEmail` this will return the first unverified email.
For `sendResetPasswordEmail` it will return the first email.

packages/server/src/email.js
```diff
@@ -0,0 +1,27 @@
...
+class Email {
+  constructor(emailConfig: Object) {
+    if (emailConfig) {
+      this.server = email.server.connect(emailConfig);
+    }
+  }
+
+  sendMail(mail: Object): Promise<Object> {
+    return new Promise((resolve, reject) => { // eslint-disable-line flowtype/require-parameter-type
+      // If no configuration for email just warm the user
```
TimMikeladze: Typo, "warn" not "warm"

packages/server/src/email.js
```diff
@@ -0,0 +1,27 @@
...
+      if (!this.server) {
+        console.log('No configuration for email, you must set an email configuration');
+        resolve();
+        return;
+      }
+      this.server.send(mail, (err: Object, message: Object) => {
```
TimMikeladze: Would be cool if user could set a function in the config to handle sending emails. At the moment I think this only supports smtp, but what about the use case when you want to do an api call to mailgun to send the email?

packages/server/src/AccountsServer.js
```diff
@@ -278,22 +294,46 @@ export class AccountsServer {
...
+    // TODO check time for expiry date
     const emailRecord = find(user.emails, (e: Object) => e.address === tokenRecord.address);
     if (!emailRecord) {
       throw new AccountsError('Verify email link is for unknown address');
     }
     await this.db.verifyEmail(user.id, emailRecord);
   }
+
+  async resetPassword(token: string, newPassword: string): Promise<void> {
+    const user = await this.db.findUserByResetPasswordToken(token);
```
pradel: All our find are with `User`even `findUserByEmail`

packages/server/src/email.js
pradel: Yeah good idea 👍

packages/server/src/AccountsServer.js
TimMikeladze: oops nevermind then 😄

pradel:
@TimMikeladze If you can review the README

TimMikeladze:
@pradel Looks good

ibelarouci:
you must modify the resolver file  of mutaion function
sendVerificationEmail
sendResetEmail and other functions to return a Boolean (true value) if the operation is succeed 
by adding a return true in block Try or updating the last return
image

PR #64, opened 2017-02-22
Fix flow and export typings

davidyaha:


packages/server/src/emailTemplates.js
```diff
@@ -1,4 +1,16 @@
 // @flow
+import type { UserObjectType } from '@accounts/common';
+
+export type EmailTemplateType = {
+  subject: (user?: UserObjectType) => string,
```
pradel: You can add an optional `from` parameter `from?: string`

davidyaha: Will it be used by the package?

pradel: It is already implemented

TimMikeladze:
👍

PR #67, opened 2017-02-23
Set token storage

davidyaha:
Allow users to supply their own storage. Mainly for react-native users that uses [AsyncStorage](http://facebook.github.io/react-native/releases/0.41/docs/asyncstorage.html#asyncstorage) which is a superset of the borwser's `localStorage`

### DIFF
--- a/packages/client/src/config.js
+++ b/packages/client/src/config.js
@@ -7,6 +7,7 @@ export default {
   store: null,
   reduxLogger: null,
   reduxStoreKey: 'accounts',
+  tokenStorage: global.localStorage,
   server: '',
   localStoragePrefix: '',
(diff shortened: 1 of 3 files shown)


packages/client/src/config.js
```diff
@@ -7,6 +7,7 @@ export default {
   store: null,
   reduxLogger: null,
   reduxStoreKey: 'accounts',
+  tokenStorage: global.localStorage,
   server: '',
   localStoragePrefix: '',
```
TimMikeladze: Maybe rename `localStoragePrefix` to `tokenStoreagePrefix` to be more inline with the naming scheme.

PR #69, opened 2017-02-24
Throw error if no tokens are set when resuming session

TimMikeladze:


davidyaha:
This PR is a WIP? Just needs to fix the test if I understand correctly

TimMikeladze:
Tests fixed.

PR #70, opened 2017-02-24
Add password recovery functions to client

pradel:
#40

### DIFF
--- a/packages/client/src/AccountsClient.js
+++ b/packages/client/src/AccountsClient.js
@@ -226,6 +226,38 @@ export class AccountsClient {
       throw new AccountsError(err.message);
     }
   }
+
+  async verifyEmail(token: string): Promise<void> {
+    try {
+      await this.transport.verifyEmail(token);
+    } catch (err) {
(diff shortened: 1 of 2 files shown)


packages/client/src/AccountsClient.js
```diff
@@ -269,6 +301,18 @@ const Accounts = {
   refreshSession(): Promise<void> {
     return this.instance.refreshSession();
   },
+  verifyEmail(token: string): Promise<void> {
+    return this.instance.verifyEmail(token);
+  },
+  resetPassword(token: string, newPassword: string): Promise<void> {
+    return this.instance.resetPassword(token, newPassword);
+  },
+  requestPasswordReset(userId: string, email?: string): Promise<void> {
```
TimMikeladze: Implementation question: Is it necessary to have `userId` passed in here? This code is being called on the client who isn't signed in, where does the userId come from?

TimMikeladze:
Is this ready to merge?

pradel:
I will add some tests before

pradel:
Should be good now

PR #71, opened 2017-02-25
bugfix(async-storage): added support for async storage

dotansimha:


packages/client/src/AccountsClient.js
```diff
@@ -73,11 +73,14 @@ export class AccountsClient {
...
   }
 
-  tokens(): TokensType {
-    return {
-      accessToken: this.storage.getItem(getTokenKey(ACCESS_TOKEN, this.options)),
-      refreshToken: this.storage.getItem(getTokenKey(REFRESH_TOKEN, this.options)),
-    };
+  async tokens(): Promise<TokensType> {
+    return Promise.all([
+      Promise.resolve(this.storage.getItem(getTokenKey(ACCESS_TOKEN, this.options))),
```
davidyaha: I think that setItem and remove is async as well on AsyncStorage. Maybe write a wrapper class that hides any storage that is sync for all methods we use?

dotansimha: Done :) Thanks

PR #72, opened 2017-02-25
Improved the email and username validators

davidyaha:


davidyaha:
I think custom validators in any form is needed. Let's open an issue on this one?

PR #77, opened 2017-03-06
chore(build): added support for UMD build along with ES6 build

dotansimha:
Also, added loginWithPassword result to callback and as return value

### DIFF
--- a/packages/client/webpack.config.js
+++ b/packages/client/webpack.config.js
@@ -21,6 +21,14 @@ module.exports = {
         test: /\.js$/,
         loader: 'babel',
         exclude: /node_modules/,
+        query: {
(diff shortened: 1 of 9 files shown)


packages/client/webpack.config.js
```diff
@@ -21,6 +21,14 @@ module.exports = {
         test: /\.js$/,
         loader: 'babel',
         exclude: /node_modules/,
+        query: {
+          babelrc: false,
```
TimMikeladze: Why not reuse the existing .babelrc files?

dotansimha: Because we need a different config with different plugins and presets for CommonJS build

PR #78, opened 2017-03-08
Added typings to configuration objects on all three packages

### LINKED ISSUE #29: Add flow typings for configs

davidyaha:
Closes #29.


packages/client/src/AccountsClient.js
```diff
@@ -264,6 +270,7 @@ export class AccountsClient {
   }
 
   async requestPasswordReset(email?: string): Promise<void> {
```
pradel: flow email is optional must be `email: string`

packages/client/src/AccountsClient.js
```diff
@@ -272,6 +279,7 @@ export class AccountsClient {
   }
 
   async requestVerificationEmail(email?: string): Promise<void> {
```
pradel: same

packages/client/src/AccountsClient.js
davidyaha: This is how it was, but come to think of it, changing it basically means nothing :)
Although I would like the email to be validated before making a network call.

davidyaha: This is how it was, but come to think of it, changing it basically means nothing :)
Although I would like the email to be validated before making a network call.

davidyaha: This is how it was, but come to think of it, changing it basically means nothing :)
Although I would like the email to be validated before making a network call.

pradel: Yeah I was doing that because meteor was taking a optional argument. But from the client the email should be mandatory I think, my mistake :)

packages/client/src/AccountsClient.js
```diff
@@ -202,7 +201,11 @@ export class AccountsClient {
       await this.storeTokens(res);
       this.store.dispatch(setTokens(res.tokens));
       this.store.dispatch(setUser(res.user));
-      this.options.onSignedInHook();
+
+      if (this.options.onSignedInHook) {
```
TimMikeladze: Maybe check to make sure it's a function for the sake of consistency?

packages/server/src/AccountsServer.js
```diff
@@ -507,7 +510,9 @@ export class AccountsServer {
     address = this._getFirstUserEmail(user, address); // eslint-disable-line no-param-reassign
     const token = generateRandomToken();
     await this.db.addResetPasswordToken(userId, address, token);
-    const resetPasswordUrl = `${this._options.siteUrl}/reset-password/${token}`;
+
+    const siteUrl = this._options.siteUrl || config.siteUrl;
+    const resetPasswordUrl = `${siteUrl}/reset-password/${token}`;
```
TimMikeladze: Should use resetPasswordPath from the config instead of `reset-password`

davidyaha: Probably not my code to change.. @pradel Do you care if change that?

pradel: No problem :)

davidyaha: @TimMikeladze @pradel Actually, this path this is a config from the client so I rather have that change in another PR

PR #79, opened 2017-03-08
Tests for sendEnrollementEmail and _getFirstUserEmail

TimMikeladze:


packages/server/src/AccountsServer.spec.js
```diff
@@ -748,5 +748,81 @@ describe('Accounts', () => {
...
+        try {
+          await Accounts.sendEnrollmentEmail('userId', 'email');
+          throw new Error();
+        } catch (err) {
+          expect(err.message).toEqual('User not found');
+        }
+      });
+      it('adds email verification token and sends mail', async () => {
+        const addResetPasswordToken = jest.fn();
+        const _getFirstUserEmail = jest.fn(() => 'user@user.com');
```
pradel: I think you don't need to create a mock for `_getFirstUserEmail` here

PR #84, opened 2017-03-12
Added my modular authorization notes

davidyaha:
I had a hard time making sense out of things without getting it all written.
Would love to hear feedback on my assumptions and maybe things I did not understand correctly.
The notes might be a little messy but I tried to format it in some way. Hope it's understandable.

### DIFF
--- a/res/notes/modular-auth.md
+++ b/res/notes/modular-auth.md
@@ -1,0 +1,131 @@
+## Modular Auth notes
+
+- Authentication service
+    - Each service will have it's own repository to hold implementation
+    
+    - Things to consider
+        - ui
+            - Forms or any view component should be defined for each component.
+            - Idealy accounts/client will passed on the ui config method and it will then be used by the ui components 
+        - data store
+            - Store auth details for document based stores
+                - How indexes are handled?
+            - map to relational entries (for sql implmentation)
+        - added endpoints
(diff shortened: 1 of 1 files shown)


res/notes/modular-auth.md
```diff
@@ -0,0 +1,131 @@
...
+            - should expose the express/koa/hapi middleware if needed.
+    
+    - Register service plugins
+        - Register on client
+        - Register on server
+        - Suggested API
+        ```typescript
+        // client code
+        const accountsClient = new AccountsClient({
+            authServices: {
```
TimMikeladze: Another approach is to have the client do a query to the server to determine which services have been configured.

davidyaha: It's possible but most services (especially oauth) will need some configurations on the client side as well so it's probably easier to the let the developer configure it's own instance and supply it to the `accountsClient`.
Also I think it will easier for developers to write client with other programming languages that use an AccountsServer service.

pradel: With this method it will take more time to load the page also

davidyaha: You mean by doing a network call to the server?

pradel: yep

PR #85, opened 2017-03-13
Improved email configuration typings

davidyaha:
LMK if you want to add the emailjs config types as well


pradel:
Maybe just a link to emailjs config is enough ?

PR #86, opened 2017-03-13
feat(tokens): eager load of tokens from storage

dotansimha:
The goal is to load from the storage only once, when the app loads, and then use the inner redux store to manage the token in-memory. 
This way, we load the tokens eagerly from the storage when the `AccountsClient.config` called, and store them in-memory.
Also, the `tokens()` is always sync now, even if the storage is async (like with `AsyncStorage` of react-native)

### DIFF
--- a/packages/client/src/AccountsClient.js
+++ b/packages/client/src/AccountsClient.js
@@ -78,15 +78,26 @@ export class AccountsClient {
     return Promise.resolve(this.storage.removeItem(keyName));
   }
 
+  async loadTokensFromStorage(): Promise<void> {
+    const tokens = {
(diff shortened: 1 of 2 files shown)


packages/client/src/AccountsClient.js
```diff
@@ -295,11 +306,13 @@ export class AccountsClient {
 const Accounts = {
   instance: AccountsClient,
   ui: {},
-  config(options: AccountsClientConfiguration, transport: TransportInterface) {
+  async config(options: AccountsClientConfiguration, transport: TransportInterface): Promise<any> {
```
davidyaha: `Promise<AccountsClient>`

dotansimha: done

PR #87, opened 2017-03-14
Added onUserCreated and validateNewUser functions to config

TimMikeladze:


packages/client/src/AccountsClient.spec.js
```diff
@@ -151,6 +151,27 @@ describe('Accounts', () => {
       expect(Accounts.instance.loginWithPassword.mock.calls[0][0]).toEqual({ id: '123' });
       expect(Accounts.instance.loginWithPassword.mock.calls[0][1]).toEqual('123456');
     });
+    it('calls onUserCreated after succesfull user creation', async () => {
```
dotansimha: Maybe add a test for failing hook that throws an exception?

packages/client/src/AccountsClient.js
```diff
@@ -172,10 +172,14 @@ export class AccountsClient {
     }
 
     try {
+      const { onUserCreated } = this.options;
       const userId = await this.transport.createUser(user);
       if (callback && isFunction(callback)) {
         callback();
       }
+      if (isFunction(onUserCreated)) {
+        await onUserCreated({ id: userId });
```
dotansimha: In case of an error, you want to prevent the creation of the user? or it's just a notification hook? (if so, it should be wrapped with try/catch, so it won't fail the user creation)

packages/server/src/AccountsServer.js
```diff
@@ -176,13 +176,30 @@ export class AccountsServer {
...
-    });
+    };
+
+    let userIsValid = true;
+    if (isFunction(validateNewUser)) {
+      userIsValid = await validateNewUser(proposedUserObject);
+    }
+
+    if (!userIsValid) {
+      throw new AccountsError('User did not pass validation');
```
dotansimha: Maybe add the actual error to the exception? 
Also, maybe wrap with try/catch the call to `validateNewUser `?

davidyaha: I would much rather the validateNewUser to return a rejected promise. the code here we could do `await` and it will basically throw. Only thing to think about is whether or not the error should be wrapped as an `AccountsError`

packages/server/src/AccountsServer.js
```diff
@@ -176,13 +176,30 @@ export class AccountsServer {
...
+    }
+
+    const userId: string = await this.db.createUser(proposedUserObject);
+
+    if (isFunction(onUserCreated)) {
+      const createdUser = await this.findUserById(userId);
+      await onUserCreated(createdUser);
```
davidyaha: Should we await for this hook? what is it's purpose?

PR #93, opened 2017-03-20
Impersonation

tomermoshe:


packages/client/src/module.js
```diff
@@ -30,6 +34,9 @@ const reducer = (state = initialState, action) => {
       return state.set('tokens', Map(tokens));
     }
     case CLEAR_TOKENS: {
+      if (!state.get('isImpersonated')) {
```
davidyaha: if not impersonated, clear original tokens? If you are not impersonated there should not be tokens in the first place isn't there? And if you are impersonated, what clearing original token mean?
I think you should remove that from this reducer and handle that on `SET_IMPERSONATED` false.

packages/server/src/AccountsServer.js
```diff
@@ -205,6 +208,84 @@ export class AccountsServer {
   }
 
   /**
+   * @description Sets an impersonation rule.
+   * @param {string} userId - User id.
+   * @param {Function<boolean>} usersFilterFn - impersonation predicate.
+   * @returns {boolean} - true if user can impersonate to impersonation user.
+   */
+  setImpersonationRule(userId: string, usersFilterFn: Function<any, boolean>) {
```
davidyaha: This behaviour is not flexible enough. It means that at the stage of my server startup, I need to know which userIds are able to impersonate other users. Or rather I need to configure that every time my "list" of my privileged users changes. A much more flexible and easier approach is to allow a simple function of the following signature `(user: UserObject, impersonateToUsername: string) => Promise<boolean>` and this will potentially do whatever is needed to come up with the answer

packages/server/src/AccountsServer.js
```diff
@@ -205,6 +208,84 @@ export class AccountsServer {
...
+      throw new AccountsError('User not found');
+    }
+
+    const impersonatedUser = await this.db.findUserByUsername(username);
+    if (!impersonatedUser) {
+      throw new AccountsError(`User ${username} not found`);
+    }
+
+    const usersFilterFn = this.getImpersonationRule(user.id);
+    if (!usersFilterFn || !usersFilterFn(impersonatedUser)) {
```
davidyaha: impersonation rules should be `async` functions

packages/server/src/AccountsServer.spec.js
```diff
@@ -28,7 +28,10 @@ describe('Accounts', () => {
 
     it('set custom password authenticator', () => {
       const db = {};
-      Accounts.config({ passwordAuthenticator: () => { } }, db);
```
davidyaha: Please revert this change

PR #96, opened 2017-03-31
Make resume session with hook

BradfordMedeiros:
Tried to create some distinction between resume session and refresh session.  The on resume session hook allows you to trigger actions via the hook that indicate first time log on.  For example, a user may wish to show a prompt to the user during initial log on.  To do so, they would attach the prompt to this on resume session hook.  signedInHook works too, but only captures initial login, not when the user uses tokens to re-authenticate. 

This has a corresponding change in rest package to change authFetch to call onRefreshSession (since this would typically be called after initial authentication)

### DIFF
--- a/packages/client/src/AccountsClient.js
+++ b/packages/client/src/AccountsClient.js
@@ -177,9 +177,13 @@ export class AccountsClient {
   }
 
   async resumeSession(): Promise<void> {
-    await this.refreshSession();
-    if (this.options.onResumedSessionHook && isFunction(this.options.onResumedSessionHook)) {
-      this.options.onResumedSessionHook();
+    try {
+      await this.refreshSession();
+      if (this.options.onResumedSessionHook && isFunction(this.options.onResumedSessionHook)) {
+        this.options.onResumedSessionHook();
+      }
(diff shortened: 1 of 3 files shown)


packages/client/src/AccountsClient.js
```diff
@@ -177,8 +177,13 @@ export class AccountsClient {
   }
 
   resumeSession(): Promise<void> {
-    // TODO Should there be any additional resume session logic here?
-    return this.refreshSession();
+    const session = this.refreshSession();
```
TimMikeladze: Please use the async / await syntax to be consistent with the rest of the code.

packages/client/src/AccountsClient.js
```diff
@@ -177,8 +177,13 @@ export class AccountsClient {
...
+    session.then(() => {
+      if (this.options.onResumedSessionHook && isFunction(this.options.onResumedSessionHook)) {
+        this.options.onResumedSessionHook();
+      }
+    });
+    return session;
```
TimMikeladze: If you're returning a value here then please update the return type of the this method to not be `Promise<void>`

PR #98, opened 2017-04-02
feat(user-object-sanitizer): added user object sanitizer

dotansimha:
First, we need to sanitize the User object, right before passing it to the client or external functions to avoid exposure of `services` array with the encrypted passwords.
Also, added to `options` the ability to provide another external sanitizer, so now the developer can filter fields before sending it to the client.

### DIFF
--- a/packages/server/src/AccountsServer.js
+++ b/packages/server/src/AccountsServer.js
@@ -1,6 +1,6 @@
 // @flow
 
-import { omit, isString, isPlainObject, isFunction, find, includes, get } from 'lodash';
+import { pick, omit, isString, isPlainObject, isFunction, find, includes, get } from 'lodash';
 import EventEmitter from 'events';
 import jwt from 'jsonwebtoken';
 import {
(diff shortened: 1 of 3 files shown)


pradel:
Nice but maybe it's better to use `pick` than `omit`?

dotansimha:
@pradel, the internal uses `omit` because we just want to strip the `services` object.
The signature for the custom sanitizer is `(user, omitFunction)` so it will be easy to implement the method without any other dependencies. 
Maybe expose the `pick` as well, so the developers will be able to choose?

pradel:
With the custom function you will be able tu use `pick` in your `omitFunction` so it's okay 😄

dotansimha:
@pradel , also, I don't want to to use `pick` inside the internal sanitizer because then we might strip fields that aren't managed by `jsaccounts` from the user.

pradel:
I was thinking in a meteor way only fields in profile should be accessible on client by default

dotansimha:
Meteor exposes also the rest of the fields, not only the `profile` field. (am I wrong?)

davidyaha:
I think the direction we are starting to head towards will leave the `profile` section of a user on a separate data store or even owned by the microservice it self.
IMO profile should not be mentioned in this package at all.

davidyaha:
@TimMikeladze Are we good on this PR?

PR #101, opened 2017-04-02
Added loginWithUser to allow for any auth support

davidyaha:
After long, and hard thought (not really), that seems like the first step to modular authentication strategy. It also allows for any consuming code to have it's own auth process and setup a session for a given user.

### DIFF
--- a/packages/server/src/AccountsServer.js
+++ b/packages/server/src/AccountsServer.js
@@ -178,16 +178,7 @@ export class AccountsServer {
         throw new AccountsError('User not found', user, 403);
       }
 
-      const sessionId = await this.db.createSession(foundUser.id, ip, userAgent);
-      const { accessToken, refreshToken } = this.createTokens(sessionId);
-      const loginResult = {
-        sessionId,
-        user: foundUser,
-        tokens: {
-          refreshToken,
-          accessToken,
-        },
-      };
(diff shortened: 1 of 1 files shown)


packages/server/src/AccountsServer.js
```diff
@@ -240,6 +231,32 @@ export class AccountsServer {
...
+   * @description Server use only. This method creates a session
+   *              without authenticating any user identity.
+   *              Any authentication should happen before calling this function.
+   * @param {UserObjectType} userId - The user object.
+   * @param {string} ip - User's ip.
+   * @param {string} userAgent - User's client agent.
+   * @returns {Promise<LoginReturnType>} - Session tokens and user object.
+   */
+  // eslint-disable-next-line max-len
+  async loginWithUser(user: UserObjectType, ip?: ?string, userAgent?: ?string): Promise<LoginReturnType> {
```
dotansimha: @davidyaha what about triggering the server hooks? (`onLoginSuccess` / `onLoginError`)

dotansimha: @davidyaha what about triggering the server hooks? (`onLoginSuccess` / `onLoginError`)

davidyaha: These are handled on the using function as they are otherwise hard to control and prevent duplicate events being emitted.

PR #104, opened 2017-05-18
Changes for redesign

pradel:
- Add a loading redux prop
- Add a setup function that checks for the tokens
- Add a changeRoute function that the user provide to be able to redirect the forms

```javascript
const accountsClient = new AccountsClient(
  {
    server: 'http://localhost:3010',
    title: 'express-rest',
    loginPath: '/login',
    signUpPath: '/signup',
    homePath: '/',
    changeRoute: path => history.push(path),
  },
);
```

I still need to update the test


packages/client/src/validate.js
```diff
@@ -0,0 +1,14 @@
+import isEmpty from 'lodash/isEmpty';
+
+export default {
+  logIn(accountsClient, { user, password }) {
```
davidyaha: Having this kind of function here means we are sticking with the password method as a the main authentication method. If this is used in the react package, I am not sure really how to break it down to phases

PR #106, opened 2017-07-11
import lodash functions as modules instead of the whole package

ozsay:


dotansimha:
@ozsay 
Maybe also remove `lodash` from `package.json` and replace with something like this: https://www.npmjs.com/package/lodash.isstring ? it won't require to use tree shaking.

pradel:
I think using `lodash/:function` doesn't require tree shaking.

davidyaha:
Closed as this was done on the typescript migration.

PR #108, opened 2017-07-12
Setup typescript

pradel:
Also migrated common package


pradel:
It looks like the tests are failing on the `AccountsError`, there are some problems extending an error in typescript.

TimMikeladze:
I looked through the code a bit. My primary thought is this:

We've all been using the `@accounts` packages for a few months now in projects and have gathered information about how to improve the API and developer experience. Instead of converting the existing code to Typescript we should instead take this opportunity to revaluate the `@accounts` design and rewrite all if not most of the code. @davidyaha began some design thoughts in #84, we should improve upon this, making sure core contributors have weighed in on the overall project design before we begin writing serious code.

davidyaha:
@TimMikeladze Actually @pradel  already implemented most of the wanted changes in https://github.com/js-accounts/next

This conversion to TS is only first step towards modular auth services https://github.com/js-accounts/accounts/issues/107

TimMikeladze:
@davidyaha That's awesome! I didn't realize that repo was a thing. Great work @pradel

davidyaha:
Don't merge before we check with canary

PR #109, opened 2017-07-17
send impersonator user id field to db

stephschuster:


dotansimha:
Looks good, but we need to add to the the DB interfaces as well :)

stephschuster:
@dotansimha already added to Mongo project, exists other project that i need to update?

stephschuster:
updated on redis: https://github.com/js-accounts/redis/pull/10
updated on mongo: https://github.com/js-accounts/mongo/pull/26

Urigo:
@stephschuster !!!!   🥇

PR #110, opened 2017-07-19
Fix: invalid key in User object validation

kamilkisiela:
`PasswordLoginUserIdentityType` has username, email and id. 
`isValidUserObject()` was looking for a `user` key instead of `username`.


davidyaha:
Thanks @kamilkisiela!

PR #115, opened 2017-10-29
make AccountsServer more extensible

sloops77:


pradel:
Hi, thanks for your contribution, can you explain your change please?

sloops77:
My motivation was that it allows a change to the implementation of findUserByUsername, findUserByEmail within the AccountServer itself and have that apply to all callers whether internal or external.

PR #119, opened 2017-11-22
Fixes for reset password

dotansimha:
In some version of Meteor, the `services.password.reset` in an object, and in some an Array.
Also, in some versions the reset password token has an `address` field, and in some it's `email`. 

This PR includes fixes for both issues

### DIFF
--- a/packages/server/src/accounts-server.ts
+++ b/packages/server/src/accounts-server.ts
@@ -697,7 +697,8 @@ export class AccountsServer {
 
     // TODO move this getter into a password service module
     const resetTokens = get(user, ['services', 'password', 'reset']) || [];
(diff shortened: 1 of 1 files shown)


packages/server/src/accounts-server.ts
```diff
@@ -695,18 +696,21 @@ export class AccountsServer {
     }
 
     // TODO move this getter into a password service module
-    const resetTokens = get(user, ['services', 'password', 'reset']);
-    const resetTokenRecord = find(resetTokens, t => t.token === token);
+    const resetTokens = get(user, ['services', 'password', 'reset']) || [];
```
davidyaha: get has default value `get(user, ['services', 'password', 'reset'], []);`

PR #120, opened 2017-12-04
twitter + instagram

pradel:
This is a test in attempt to separate the oauth providers. Let me know what you think.

I splitted the accounts providers in there own packages.
Also there is a new method unlink for the oauth class that let a user unlink a service.

Example:
```js
import Oauth from '@accounts/oauth';
import OauthTwitter from '@accounts/oauth-twitter';

const oauth = new Oauth({
     instagram: new OauthInstagram(),
     twitter: new OauthTwitter({
          key: 'XX',
          secret: 'XX'
    }),
});
```


packages/oauth-twitter/src/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+class AccountsOauthTwitter {
+  private options: AccountsOauthTwitterOptions;
+
+  constructor(options: AccountsOauthTwitterOptions) {
+    this.options = options;
+  }
+
+  public authenticate(params) {
+    const oa = new oauth.OAuth(
```
davidyaha: Should maybe create the oauth object on the constructor. Also maybe use [purest?](https://github.com/simov/purest)

packages/oauth-twitter/src/index.ts
```diff
@@ -0,0 +1,52 @@
+import * as oauth from 'oauth';
```
TimMikeladze: Just to confirm: if we're using `oauth` package does this mean we can remove the `grant` dependency overall?

packages/oauth-twitter/src/index.ts
```diff
@@ -0,0 +1,52 @@
...
+        params.access_token,
+        params.access_token_secret,
+        (err, data) => {
+          if (err) {
+            reject(err);
+          } else {
+            data = JSON.parse(data);
+            const user = {
+              id: data.id_str,
+              screenName: data.screen_name,
```
TimMikeladze: What are your thoughts on returning the entirety of `data` response and allowing the developer to decide if they want to consume the additional data (part of `@accounts/server`). The only fields which are required to be returned here are some sort of unique identifier and the tokens correct?

packages/oauth-twitter/src/index.ts
```diff
@@ -0,0 +1,52 @@
+import * as oauth from 'oauth';
+
+export interface AccountsOauthTwitterOptions {
+  key: string;
+  secret: string;
+}
+
+class AccountsOauthTwitter {
```
TimMikeladze: This is a really minor comment, feel free to disregard.

For consistency, let's always use `OAuth` instead of `Oauth`. So instead of `AccountsOauthTwitterOptions` it'll be `AccountsOAuthTwitterOptions`. The reason for this is because the `oauth` library has a constructor `OAuth`.

packages/oauth-twitter/src/index.ts
pradel: Not really the tokens passed down to the oauth package came from grant. We need to verify these token return the good user, this is why we have to use the oauth module here.

packages/oauth-twitter/src/index.ts
pradel: Yeah you are right, I just took the same implementation as meteor.
Maybe we could allow a custom user function which indicate which fields should be saved?

PR #125, opened 2017-12-15
WIP - Refactoring Accounts-JS

Aetherall:
This is the work I have done from 2 weeks

consider trying it with [https://github.com/Aetherall/accounts-workbench](https://github.com/Aetherall/accounts-workbench)


Aetherall:
More Informations
--------------------------

Initially I was working on the **Token Transport**, allowing the user to choose where to store tokens based on the request object.

Then I made breaking changes on the way the accounts server used the notifications and authentications services

 - **Accounts Server**
 - Authentication 
	 - **PasswordService**
	 - **OAuthService**
		 - **FacebookProvider**
		 - **TwitterProvider**
		 - etc...
 - Authorization
	 - **TokenManager**
 - Notification
	 - Email
		 - Service
			 - **EmailServiceDebug**
			 - **EmailServiceMailgun**
		 - Plugins
	      	- **EmailPluginPassword**
 - Transport
	 - Express
		 - **ExpressTransport**
		 - TokenTransport
			 - **ExpressTTBody**
			 - **ExpressTTCookies**
			 - **ExpressTTHeaders**
	 - **TokenTransportManager**
 - Database
	 - **MongoInterface**

So...
Wtf is this


----------


### Part 1 : ExpressTransport

Let's begin with the Express Transport.

#### Send Response
First I modified the way the response is send back to the client.

Instead of doing it right in each method, I created a new private method "send" which takes 
the **res.toSend** variable and merge it with the body of the response.

I made this change to allow the Token Transports to write in the response object without interfering with the others methods results.

#### Authentication Middleware
There is a public method called middleware, which is in fact the authentication middleware.
( I should change his name ) 
It does pretty much the same as before, resuming the session and store the user in the req.user

#### Routes Middleware
The second middleware provides the routes using the functionnalities of A.js

There is four static routes :
-  ````/impersonate ````
-  ````/user````
-  ````/refreshTokens````
-  ````/logout````

Theses does pretty much the same thing: calling the corresponding methods on the AccountsServer Class and send back the response. 
> **Note:**

> Whe should probably abstract them into one dynamic route or even better merge them to the dynamic one somehow




And there is a dynamic route : ````/:service/:provider?/:action````

This one is designed to achieve these matches:

  - ````/password/register```` 
  ````javascript 
  => 
  { 
    service: 'password', 
    action: 'register'
  }
  ````
  
  - ````/oauth/facebook/authenticate```` 
  ````javascript 
  => 
  { 
    service: 'oauth',
    provider: 'facebook'
    action: 'authenticate'
  }
  ````
  
  - ````/oauth/facebook/callback```` 
  ````javascript 
  => 
  { 
    service: 'oauth',
    provider: 'facebook'
    action: 'callback'
  }
  ````

> **Note:**

> I think I should edit it to ````/:1/:2?/:3?/:4?```` 
> so we can actually make calls like ````/password/sendResetPassword/phone````

So the goal behind this is to make public functions of Authentication Services accessible through the Express Api dynamically

no more password routes on a OAuth Only app

package.json
```diff
@@ -1,37 +1,27 @@
...
-    "tslint": "^5.8.0",
-    "typescript": "^2.4.1"
-  }
+	"private": true,
+	"name": "@accounts/workbench",
+	"scripts": {
+		"initial": "yarn install && lerna bootstrap",
+		"start": "lerna exec --parallel -- yarn start",
+		"clean": "lerna exec -- yarn clean",
+		"clean_": "lerna exec -- yarn clean_",
```
TimMikeladze: What are the underscored scripts for? Naming should be more descriptive to remove ambiguity. 

Also, usually `:` is used for scripts like `link:global`.

package.json
```diff
@@ -1,37 +1,27 @@
 {
-  "private": true,
-  "scripts": {
```
TimMikeladze: Let's not delete all these scripts and repurpose them? We still need these such as the publish and test scripts.

TimMikeladze:
@Aetherall 

Reviewing this PR it looks like it removes/replaces a lot of functioning code and configuration which was already functioning fine on the `auth-services` branch and I don't needed to be changed. Since this PR does introduce additional functionality, could you please reduce the scope of it to just send that code to the `auth-services` branch in this PR. Preferably broken into separate feature commits?  

The changes to the express middleware should go to https://github.com/accounts-js/rest. Note: There's already a PR there which @pradel has sent to go along with the changes introduced by the `auth-services` PR.

package.json
Aetherall: I agree, but this part is not clear to me yet, I don't fully understand how git hooks works
If I just copy/paste the current scripts it will work ?

package.json
Aetherall: Understood

the underscore one is for removing the node_modules too
I will change this in the next commit

Aetherall:
@TimMikeladze 

One goal was to merge all the repositories into the same monorepo
It seemed cleaner to me and I think it improve the global vision of the project in development

Also the minor changes into the packages aim to clean the code, and keep consistency between packages

Does it makes sense ?

I can reduce the scope but I still think it's a pity

TimMikeladze:
@Aetherall I'm not opposed to having a single mono repo for the entire accounts suite. However the scope of this PR too large, it seems to be removing or replacing code which has already been written and tested. Furthermore there's little code style/linting. I'm having a hard time pinpointing exactly what needs to be reviewed cause it's just hundreds of additions/removals without any specific justification as to why a certain block of code was changed. @davidyaha @pradel what do you think?

Aetherall:
Sure, I didn't see it this way

Aetherall:
The work done on next-branch will be split up into multiple PRs !

PR #129, opened 2018-01-22
Add configurable email templates

davidyaha:
Found out we were missing this part.
I think the docs now are more accurate: http://accounts-js.netlify.com/docs/emails/ 😄


pradel:
Good catch missing on master but already working in the next branch ;)

PR #156, opened 2018-03-05
use random tokens instead of sessionId inside sessions

pradel:
The goal of the pr is to change the information stored in the jwt. We are currently storing the sessionId which can be predictable, so if someone succeed to get your jwt secret, he will be able to impersonate any user you have and you will not notice it.
If we switch the session id with a random token which is not predictable we remove this potential security issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,8 @@ lib/
 node_modules/
 coverage/
 npm-debug.log
-.idea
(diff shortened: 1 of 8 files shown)


pradel:
I still need to do the mongodb pull request but I would like your thought on this one :)

PR #162, opened 2018-03-08
impersonate take impersonatedUserId instead of username

### LINKED ISSUE #161: Impersonate should use userId instead of username
Right now the `impersonate` function takes an username in order to impersonate a user, we should use the user id instead.

pradel:
Close #161


davidyaha:
This is actually a requirement on one of my clients because we might not have the user id at the client side. If you need the user id approach, maybe add those as options? Anyways, the user can only be impersonated when you implement the authorization method for it.

davidyaha:
Or at least it should be the case.

pradel:
Yeah but username should only be in the password accounts no?

pradel:
so if you are using oauth or other you will not be able to impersonate ..

davidyaha:
Why? Email/username are good for public identifier for any authentication strategy I would think. Just a matter of definition I guess..

davidyaha:
I think you are correct to add the id, just think it should be an addition, along with email maybe

davidyaha:
For instance GitHub oAuth would have a username as well as Twitter

pradel:
If you register via oauth you can have some users without username and emails so in that case this function will not work

pradel:
So in order to merge this one we need the accounts-password to have an impersonate function that take the username or email

Aetherall:
Why add a impersonate function on the accounts-password ?
OAuth users can be impersonated as well
Maybe we should consider add a method like this :
````ts
function identifyUser(identity: Identity): User

interface Identity {
	userId?: string,
	username?: string,
	email?: string
}
````
So with any info of an user we can get the User object and operate

pradel:
@davidyaha @Aetherall should be good now :)

PR #165, opened 2018-03-09
Feature/2fa

pradel:
This pr add two-factor authentication to the current password package
You can take a look at the examples repository if you want to try it :)


davidyaha:
just for docs sake, I will open an issue to consider all kinds of 2fa

PR #171, opened 2018-03-12
@accounts/error : new AccountsError package with message formatting

Aetherall:
This is a first implementation of the AccountsError package.

````ts
new AccountsError('PasswordService','authenticate','no user match tokens')
// message => '[ Accounts - PasswordService ] authenticate : no user match tokens'

new AccountsError('no user match tokens')
// message => 'no user match tokens'
````

As the error code and login info is not used yet, I removed the code concerning them so we can implement it accordingly with the current state

### DIFF
--- a/packages/error/package.json
+++ b/packages/error/package.json
@@ -31,7 +31,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/js-accounts/accounts/tree/master/packages/server"
+    "url": "https://github.com/accounts-js/accounts/tree/master/packages/error"
   },
   "keywords": [
     "rest",
--- a/packages/error/package.json
+++ b/packages/error/package.json
@@ -1,0 +1,51 @@
+{
+  "name": "@accounts/error",
+  "version": "0.1.0-beta.3",
+  "description": "Accounts-js Error",
+  "main": "lib/index.js",
+  "typings": "lib/index.d.ts",
+  "publishConfig": {
+    "access": "public"
+  },
+  "scripts": {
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: packages/error/package.json [hunk 1]
```diff
@@ -0,0 +1,51 @@
...
+    },
+    "testRegex": "(/__tests__/.*|\\.(test|spec))\\.(ts|tsx)$",
+    "moduleFileExtensions": [
+      "ts",
+      "js"
+    ]
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/js-accounts/accounts/tree/master/packages/server"
```
davidyaha: https://github.com/js-accounts/accounts/tree/master/packages/error

Aetherall: my bad, will be fixed in the next minutes

pradel:
@Aetherall is there a way to get 100% test coverage for the error package in oder to not loose `1.78%` of code coverage for the global repository?

Aetherall:
Actually the coverage missing comes from

````ts
if (typeof Error.captureStackTrace === 'function') {
  Error.captureStackTrace(this, this.constructor);
} else { 
  this.stack = (new Error(message)).stack; 
}
````

It would be a string comparison of something that is not predictable
Maybe there is an other way to do it I am not aware of ?

Aetherall:
Can I merge This ?

pradel:
You will need to prettier before I think and let's wait for @davidyaha review

davidyaha:
just figure out the tests but that is good to go IMO

PR #173, opened 2018-03-12
@accounts/types : add a types package which contains common types

Aetherall:


pradel:
Why do we need a new types package, can this go to common instead?

Aetherall:
In fact the common package actually is a dependency in some packages and a devDependency in others
the common package exports utilities that will no longer need to be shared across the suite in the future versions of Accounts-Js

That's why I am creating @accounts/types and @accounts/error

@accounts/error will be the dependency
@accounts/types will be the devDependency

It keeps the relation between packages simple

packages/types/package.json
```diff
@@ -0,0 +1,51 @@
...
+    },
+    "testRegex": "(/__tests__/.*|\\.(test|spec))\\.(ts|tsx)$",
+    "moduleFileExtensions": [
+      "ts",
+      "js"
+    ]
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/js-accounts/accounts/tree/master/packages/server"
```
davidyaha: https://github.com/js-accounts/accounts/tree/master/packages/types

Aetherall: Oh yes I am changing that

Aetherall:
There is also DatabaseInterface because it makes more sense to have it here
I am adding the readme

davidyaha:
@Aetherall A user with TS in his project needs to add this package to devDependencies as well?

Aetherall:
@davidyaha only if he is using those types I suppose...

packages/types/readme.md
```diff
@@ -0,0 +1,8 @@
+# Accounts Types
+### @accounts/types
+
+This package is a devDependency for most packages of the suite.
+
+It contains the types used across the Accouts Suite.
+
+#### Usage : This package is used internally by the suite. It's not needed to install it
```
davidyaha: OK, so maybe remove that.

Aetherall: Done

packages/types/src/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+
+export { EmailRecord } from './types/email-record'
+
+// ? Operation Results
+export { LoginResult as LoginReturnType } from './types/login-result'
+export { ImpersonationResult as ImpersonateReturnType } from './types/impersonation-result'
+
+
+// ? Operation Parameters
```
pradel: we will need to figure this before merging no?

packages/types/src/types/user.ts
```diff
@@ -0,0 +1,10 @@
+import { EmailRecord } from './email-record';
+
+export interface User {
+  username?: string;
+  email?: string;
```
pradel: is this line accurate?

Aetherall: In fact I just copied the types and renamed it

pradel: Yeah I now but was wondering if this is used or not ^^

pradel: I think it should not be used no?

packages/types/src/index.ts
Aetherall: Yes actually I want to discuss it ASAP

packages/types/src/types/user.ts
Aetherall: you mean the email field?

pradel: email string field

Aetherall: I agree, this was on the UserObjectType interface

packages/types/src/index.ts
pradel: So this will be used on the password server and password client right?

Aetherall: the @accounts/types package will be used on pretty much every @accounts package, as a devDependency

pradel: I mean the type

Aetherall: Yes it is

davidyaha: What is the idea of Operation Parameters?

Aetherall: For several operations, the function expect one object as parameters which contains all the operation needs to be complete

PR #174, opened 2018-03-12
Feature/database-manager : Allow sessions and users to be stored in different databases 

Aetherall:


Aetherall:
I think it would be a good idea if we were not planning on extend accounts-js as we want to... If we are going this direction, the accountsServer will get quickly messy, and we will need to take care of the difference in all authservices.
By having a separate package, maybe it will need to be changed too whenever we decide to change the DatabaseInterface related packages, but we will not need to change the authServices.
And in the current setup, it is easy to update that one, it took me 30s to make it compatible with two factor ( copy paste the setService method and change setService to unsetService, then do the same in the tests )

TimMikeladze:
> I think it would be a good idea if we were not planning on extend accounts-js as we want to...

> If we are going this direction, the AccountsServer will get quickly messy, and we will need to take care of the difference in all Authservices.

Can you provide me a real world example of where having the database manager would be beneficial? I'm failing to understand where exactly it improves things as opposed to handling the same logic within `AccountsServer` with far less code.


> Also, what about the case where we use a single database ? Handling this in AccountsServer would result on many additions

Handling a single database would work like this:

```javascript

const accounts = new AccountsServer({
	db: new Mongo
});

// inside AccountsServer constructor

constructor(options) {
    // improve logic and validation, just an example
	if (options.db instanceof DatabaseInterface) {
      this.db.sessions = options.db;
      this.db.users = options.db;
    } else if (options.db.sessions && options.db.users) {
       this.db.sessions = this.db.sessions;
       this.db.users = this.db.users;
    }
}

```

Aetherall:
Both methods would work the same in the end, but I made this package so we don't have to handle those cases in all future packages

the setStore function would still work on the AuthServices, and no need to rewrite packages

With the database manager, the usage of two database is transparent for the AccountsServer as well as for the AuthenticationServices

It is a matter of point of view, I prefer the splitting to be transparent to the rest of the suite, but I guess we should ask for @davidyaha and @pradel opinions

TimMikeladze:
Alright, I'm on board. @Aetherall good to merge?

Aetherall:
Yep, good to me !

PR #178, opened 2018-03-13
WIP : @accounts/token-manager Adds a package to manage jwt's

Aetherall:
This package has the purpose to be used by the AccountsServer class to manage JWT's in a cleaner way

Usage: 
````ts
// use default config
	secret: 'secret',
})

// Use custom config
	secret: 'secret',
	emailTokenExpiration: 1000 * 60,
	access: {
		algorithm: 'HS256',
		expiresIn: '10m'
	},
	refresh: {
		algorithm: 'HS256',
		expiresIn: '1d'
	}
})
````

By now it is not implemented anywhere, but the goal is to pass it to accountsServer this way :

````ts
const accountsServer = new AccountsServer({
	tokenManager
})
````


Aetherall:
Oops seems like I messed up my commits

PR #179, opened 2018-03-13
WIP : @accounts/token-manager Adds a package to manage jwt's

Aetherall:
This package has the purpose to be used by the AccountsServer class to manage JWT's in a cleaner way

Usage: 
````ts
// use default config
	secret: 'secret',
})

// Use custom config
	secret: 'secret',
	emailTokenExpiration: 1000 * 60,
	access: {
		algorithm: 'HS256',
		expiresIn: '10m'
	},
	refresh: {
		algorithm: 'HS256',
		expiresIn: '1d'
	}
})
````

By now it is not implemented anywhere, but the goal is to pass it to accountsServer this way :

````ts
const accountsServer = new AccountsServer({
	tokenManager
})
````

### DIFF
--- a/packages/token-manager/src/token-manager.ts
+++ b/packages/token-manager/src/token-manager.ts
@@ -1,0 +1,26 @@
+import { randomBytes } from 'crypto';
+import * as jwt from 'jsonwebtoken';
+import { error } from 'util';
+
+import { Configuration } from './types/configuration';
+import { TokenGenerationConfiguration } from './types/token-generation-configuration';
+
+const defaultTokenConfig: TokenGenerationConfiguration = {
+  algorithm: 'HS256',
+};
+
+const defaultAccessTokenConfig: TokenGenerationConfiguration = {
+  expiresIn: '90m',
+};
+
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: packages/token-manager/src/token-manager.ts [hunk 1]
```diff
@@ -0,0 +1,68 @@
...
+
+  private secret: string;
+
+  private emailTokenExpiration: number;
+
+  private accessTokenConfig: TokenGenerationConfiguration;
+
+  private refreshTokenConfig: TokenGenerationConfiguration;
+
+  constructor(config: Configuration) {
```
TimMikeladze: what are your thoughts on just defining and exporting `Configuration` interface within this file and eliminating `./types` entirely?

Aetherall: I want to enforce a pattern which has to be respected by all the packages of the suite, so if a potential contributor need to add a package, he will apply the pattern to his package naturally

Aetherall:
Closed in favor of #222

PR #182, opened 2018-03-13
Prettier: allow code to extend beyond 80 chars

Aetherall:
Small commit to change the prettier print-width


Aetherall:
I updated the prettier printWidth and formatted the suite
Scince this is a change accepted in `@accounts/tslint-config-accounts`, can I merge this one ?

pradel:
For me 130 char is too big I am working on a mackbook 13' and the line is bigger than my screen I would allow 100 max, but I am still in favor to keep to 80 ^^

Aetherall:
I see, 100 would go but 80 is litterally less than the half of a normal screen...

packages/password/__tests__/__snapshots__/accounts-password.ts.snap
```diff
@@ -8,6 +8,8 @@ Array [
 ]
 `;
 
+exports[`AccountsPassword authenticate return user 1`] = `"Cannot read property 'services' of undefined"`;
```
pradel: Looks like something going wrong there

package.json
```diff
@@ -12,7 +12,7 @@
     "publish:canary": "yarn run publish -- --canary",
     "test": "yarn lint && yarn compile && yarn coverage",
     "testonly": "lerna run testonly",
-    "format": "prettier --write \"packages/*/{src,__tests__}/**/*.ts\"",
+    "format": "prettier --print-width 100 --write \"packages/*/{src,__tests__}/**/*.ts\"",
```
pradel: please use the prettier config in package.json instead of a cli argument here

packages/password/__tests__/__snapshots__/accounts-password.ts.snap
Aetherall: Should be OK now

PR #186, opened 2018-03-14
Cleanup/structure: apply structure pattern to the suite

Aetherall:


Aetherall:
@davidyaha what did we said about this PR ?
Is it good to merge ?

PR #194, opened 2018-03-16
@accounts/types : export new interface names

Aetherall:
So we can move on with the new names as well


pradel:
Would be cool to use them in order to merge this one no?

Aetherall:
yes sure ! I can do this now

Aetherall:
I can merge it to this branch first, so I can make the `@accounts/types` package used by the others, what do you think @pradel ?

packages/types/src/index.ts
```diff
@@ -1,20 +1,50 @@
-export { DatabaseInterface as DBInterface } from './types/database-interface';
-import { ConnectionInformations as ConnectionInformationsType } from './types/connection-informations';
+import { DatabaseInterface } from './types/database-interface';
```
TimMikeladze: You can express importing and exporting like this. 

```javascript
`export * from './types/database-interface`
```

This will export all named export within `./types/database-interface` and you no longer need to explicitly import.

Aetherall: I did this at the moment so i can rename the interfaces on the fly, but I guess it can be written with stars in #197

packages/types/src/types/authentication-service.ts
```diff
@@ -0,0 +1,11 @@
+import { User } from './user';
+import { DatabaseInterface } from './database-interface';
+
+// import AccountsServer from '@accounts/server';
```
pradel: Not possible to do that? would be better for type checking

Aetherall: I didn't succeed... it makes a circular dependency and the suite can't compile

pradel: Okay let's keep it like that for now we can still fix it later

pradel: Can you just add a TODO as a reminder and explain the pb as comments?

PR #197, opened 2018-03-17
@accounts/types Integration to the suite

Aetherall:
This PR does use the #186 PR

This contains the @accounts/types package integrated with the others packages on the repo (except for the client)

See #194


Aetherall:
Failing tests are due to "yarn not found"
any idea why ? :/

TimMikeladze:
Hi @Aetherall, I reran the tests in the circleci dashboard and they are passing now 👍. Don't know what was wrong.

This current PR is based off of another pending PR https://github.com/accounts-js/accounts/pull/186 ? If this is correct, please edit this PR to be sent to that branch instead of master to make reviewing easier.

Aetherall:
Okay sure !

packages/database-mongo/package.json
```diff
@@ -33,15 +33,10 @@
   "author": "Leo Pradel",
   "license": "MIT",
   "devDependencies": {
-    "@accounts/common": "^0.1.0-beta.5",
-    "@accounts/server": "^0.1.0-beta.5",
+    "@accounts/types": "^0.1.0-beta.5",
```
pradel: should this be into dependencies?

packages/server/src/types/accounts-server-options.ts
```diff
@@ -0,0 +1,33 @@
...
+  userObjectSanitizer?: UserObjectSanitizerFunction;
+  impersonationAuthorize?: (
+    user: User,
+    impersonateToUser: User
+  ) => Promise<any>;
+  resumeSessionValidator?: ResumeSessionValidator;
+  siteUrl?: string;
+  prepareMail?: PrepareMailFunction;
+  sendMail?: SendMailType;
+  // https://github.com/eleith/emailjs#emailserverconnectoptions
```
pradel: I think we can remove this line no? The email option is not needed anymore since we now use the sendMail property.

packages/types/src/index.ts
```diff
@@ -10,41 +10,45 @@ import { LoginResult } from './types/login-result';
 import { ImpersonationResult } from './types/impersonation-result';
 import { Login } from './types/login';
 import { HookListener } from './types/hook-listener';
+import { AuthenticationService } from './types/authentication-service'
+
+export { AuthenticationService }
+// export { AuthenticationService as AuthService }
```
pradel: If we use the new names we can remove the comments :)

pradel:
The nodejs image was broken so not your fault :)

Looks good to me good job!

packages/database-mongo/package.json
Aetherall: I guess scince it's not hosted on the DefinitelyTyped repo, We need to have it in devDeps so it cant be resolve for compilation ? What do you think?

pradel: For me it looks like when we will install an accounts-js module sice this one is not in dependencies it will not be installed at all so typescript will fail because it will not find the type no? (I am not sure I think we will have to try this one)

packages/server/src/types/accounts-server-options.ts
Aetherall: I removed it :)

Aetherall:
Okay I see!
Thanks :D

packages/server/src/types/accounts-server-options.ts
pradel: Can you also remove the comment ? ^^

Aetherall: done !

PR #199, opened 2018-03-19
New hook system

pradel:
- Use [Emittery](https://github.com/sindresorhus/emittery) instead of EventEmitter, this will allow us to wait for the emitters during the functions
- Remove all the `onXXX` functions so user will only use the `.on` function

For example so following code will prevent users with some invalid ip to connect to your application.
```javascript
accountsServer.on('Login', async ({ connection }) => {
  if (!await isValid(connection.ip)) {
    throw new Error('User not allowed')
  }
});

// Can also be used to track services events
accountsServer.on('password.resetPassword', () => {
  // Custom logic
});
```

### DIFF
--- a/packages/server/src/utils/server-hooks.ts
+++ b/packages/server/src/utils/server-hooks.ts
@@ -1,4 +1,5 @@
 export const ServerHooks = {
+  Login: 'Login',
   LoginSuccess: 'LoginSuccess',
   LoginError: 'LoginError',
   LogoutSuccess: 'LogoutSuccess',
--- a/packages/server/__tests__/account-server.ts
+++ b/packages/server/__tests__/account-server.ts
@@ -6,6 +6,7 @@ import {
   hashPassword,
   verifyPassword,
 } from '../src/utils/encryption';
+import { ServerHooks } from '../src/utils/server-hooks';
 
 describe('AccountsServer', () => {
   const db = {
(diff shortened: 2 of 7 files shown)


packages/server/src/utils/server-hooks.ts
```diff
@@ -1,4 +1,5 @@
 export const ServerHooks = {
+  Login: 'Login',
```
pradel: Not sure about the name do you have a better Idea?

pradel:
@davidyaha @TimMikeladze @Aetherall can I have your thought on this?

packages/server/src/utils/server-hooks.ts
Aetherall: I think this is fine for now, let's see if we can find a logic name scheme when more hooks will be added

Aetherall:
The whole idea seems great to me, plus the AccountsServer class is more readable

Maybe invest how we can change the behavior of a fonction through the hooks

```ts
accountsServer.on('createUser', ({ user }) => {
  throw {...user, role: ['default']}
});
```
But it's just an idea, and probably irrelevant here

pradel:
Don't get what this example should do?

Aetherall:
This override the user object before it is inserted in the database

pradel:
@Aetherall I think this should be a real function rather than a hook

pradel:
We will need to update the docs after this pr
And we will also need to add hooks to the password package.

PR #221, opened 2018-04-01
Fix - Database Manager should use @accounts/types

Aetherall:


packages/database-manager/package.json
```diff
@@ -43,7 +41,7 @@
   "author": "Elies Lou (Aetherall)",
   "license": "MIT",
   "devDependencies": {
-    "@accounts/server": "^0.1.0-beta.3",
+    "@accounts/types": "^0.1.0-beta.10",
```
pradel: @accounts/types should be inside dependencies

Aetherall: Okay It's fixed now

PR #222, opened 2018-04-03
Token manager Implemented

Aetherall:
Here is the PR to implement the token manager with the rest of the suite

I had to open a new one because jest broke when I tried to merge with master


packages/server/src/accounts-server.ts
```diff
@@ -44,6 +32,7 @@ const defaultOptions = {
 
 export class AccountsServer {
   public options: AccountsServerOptions;
+  public tokenManager: any;
```
pradel: if you can add real typings there would be great

packages/token-manager/src/token-manager.ts
```diff
@@ -0,0 +1,67 @@
...
+    return !tokenRecord || Number(tokenRecord.when) + this.emailTokenExpiration < Date.now();
+  }
+
+  public decodeToken(token: string, ignoreExpiration: boolean = false): string | object {
+    return jwt.verify(token, this.secret, { ignoreExpiration });
+  }
+
+  private validateConfiguration(config: Configuration): void {
+    if (!config) {
+      throw new Error('[ Accounts - TokenManager ] configuration : A configuration object is needed');
```
pradel: can you use the error module?

Aetherall:
```js

import TokenManager from '@accounts/token-manager'

	secret: 'secret',
	emailTokenExpiration: 1000 * 60 * 60 * 2,
	access: {
		expiresIn: '1h'
	},
	refresh: {
		algorithm: 'HS256'
	}
})

new AccountsServer({
	db,
	tokenManager
}, {
	password: new Password(passwordConfig)
})
```

packages/token-manager/src/token-manager.ts
Aetherall: sure ! Should be done now

packages/server/src/accounts-server.ts
Aetherall: sure ! Should be done now

pradel:
So the user will have to install a new package manually, can't we just make the initialisation internal?

PR #316, opened 2018-05-02
Feature/refactor client v0.1 and add client-password package

pradel:
- [x] Refactor the client so we don't need a lot of dependencies
- [x] Create a new `client-password` package
- [x] `client-password` should encrypt password on the client side for all password related operations: `createUser`, `changePassword`, `resetPassword`
- [x] The user need to be authenticated for some functions eg `changePassword`
- [x] Tests working

The example is updated on another branch if you want to take a look https://github.com/accounts-js/examples/pull/141.

### Usage:

```javascript
const accountsClient = new AccountsClient({ ...options }, transport);
const accountsPassword = new AccountsClientPassword(accountsClient, { ...options });
```

The user should now use the `client-password.createUser` function that will encrypt the user password before sending it to the server.

The user should now use the `client-password.login` function that will encrypt the user password before sending it to the server.

By default we will only hash the password with the `SHA256` algorithm. And there is an option for the user to define the method of his choice, in that way we do not import the whole `crypto-js` library which is pretty big.

### Breaking changes

- For the users created before this version you will not be able to use the `client-password.login` that will encrypt the user password before sending it to the server.
There is two workaround:
  - continue to use the `client.loginWithService` method which does not encrypt the password.
  - inside `client-password` provide the `hashPassword` function that will just return the string in plain text for example:

```javascript
const accountsPassword = new AccountsClientPassword(accountsClient, {
  hashPassword: (password) => password
});
```


Aetherall:
Good job ! :+1:

PR #317, opened 2018-05-03
Pass some context to the login hooks

pradel:
- `ServerHooks.LoginSuccess`
- `ServerHooks.LoginError`

With this change the 2 hooks will receive the following infos:
```
{
// The service name, such as “password” or “twitter”.
service
// The connection informations <ConnectionInformations>
connection
// The user object if set
user
}
```

I also renamed the `ServerHooks.Login` to `ServerHooks.ValidateLogin` because we let the user the ability to validate the current user on the login flow.

### DIFF
--- a/packages/server/src/accounts-server.ts
+++ b/packages/server/src/accounts-server.ts
@@ -90,29 +90,38 @@ export class AccountsServer {
     params,
     infos: ConnectionInformations
   ): Promise<LoginResult> {
-    if (!this.services[serviceName]) {
-      throw new Error(`No service with the name ${serviceName} was registered.`);
-    }
-    const user: User = await this.services[serviceName].authenticate(params);
-    if (!user) {
-      throw new Error(`Service ${serviceName} was not able to authenticate user`);
-    }
-
-    await this.hooks.emitSerial(ServerHooks.Login, {
+    const hooksInfo: any = {
       // The service name, such as “password” or “twitter”.
       service: serviceName,
-      // The user object
-      user,
       // The connection informations <ConnectionInformations>
       connection: infos,
-    });
-    return this.loginWithUser(user, infos);
+    };
+    try {
+      if (!this.services[serviceName]) {
+        throw new Error(`No service with the name ${serviceName} was registered.`);
+      }
+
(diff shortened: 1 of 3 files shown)


packages/server/src/utils/server-hooks.ts
```diff
@@ -1,5 +1,5 @@
 export const ServerHooks = {
-  Login: 'Login',
+  ValidateLogin: 'ValidateLogin',
```
davidyaha: This is a breaking change.. until we're ready to break the API, we can deprecate and support both names

packages/server/src/accounts-server.ts
```diff
@@ -90,60 +90,59 @@ export class AccountsServer {
...
-   *              Any authentication should happen before calling this function.
+   * @description Server use only.
+   * This method creates a session without authenticating any user identity.
+   * Any authentication should happen before calling this function.
    * @param {User} userId - The user object.
    * @param {string} ip - User's ip.
    * @param {string} userAgent - User's client agent.
    * @returns {Promise<LoginResult>} - Session tokens and user object.
    */
   public async loginWithUser(user: User, infos: ConnectionInformations): Promise<LoginResult> {
```
davidyaha: As this is a public method, some users that call loginWithUser directly (maybe even @dotansimha?) will not have the login and error hooks called. This is a breaking change, are we ready for minor release?

packages/server/src/utils/server-hooks.ts
pradel: Yeah I agree will do the change

packages/server/src/accounts-server.ts
pradel: Hum true, I was wondering is this function should trigger or not the hooks since it can only be called server side?

davidyaha: Yeah I think it is currently one of the more flexible ways to create your own authentication methods outside of our auth-services mechanism. I don't mind simplifying this code, but it is breaking..

pradel: So the problem is that we currently need to have more info inside the hooks can you think about a better approach?

PR #318, opened 2018-05-03
create a types folder

pradel:
- create a types folder to align to the rest of the code
- create a `getUserTwoFactorService` utils that can be used to check if the user have two factor enabled or not

### DIFF
--- a/packages/two-factor/src/two-factor.ts
+++ b/packages/two-factor/src/two-factor.ts
@@ -1,18 +1,8 @@
 import * as speakeasy from 'speakeasy';
 import { User, DatabaseInterface } from '@accounts/types';
 import { errors } from './errors';
-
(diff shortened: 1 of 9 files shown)


packages/two-factor/src/two-factor.ts
```diff
@@ -50,7 +33,7 @@ export class TwoFactor {
       throw new Error(errors.codeRequired);
     }
 
-    const twoFactorService = this.getUserService(user);
+    const twoFactorService = getUserTwoFactorService(user);
```
davidyaha: Do we use the data from the twoFactor service? Maybe change `getUserTwoFactorService` to `hasTwoFactorService(user: User): boolean`

davidyaha: okay 👍

PR #323, opened 2018-05-09
WIP Feature/database redis

pradel:


pradel:
Introduced a new `DatabaseInterfaceSessions` interface that contain only the session functions. Like that we can create only session store.

pradel:
@TimMikeladze still okay with the latest changes?

TimMikeladze:
@pradel 👍 Looks good.

PR #331, opened 2018-05-15
No longer set profile data in OAuth

kamilkisiela:
@dotansimha


dotansimha:
BTW, ci is only failing because of a decrease in coverage....

TimMikeladze:
@dotansimha That sounds good to me.

PR #345, opened 2018-06-06
Feature/remove common package

pradel:
Remove the common package since we don't use it anymore.
No breaking changes expected if you don't rely on it.


pradel:
@davidyaha we will need to use the new accounts-error package but will do it later in other pr :)

PR #346, opened 2018-06-09
Update TypeScript and fix emittery issue

kamilkisiela:
With TypeScript 2.9.1, current `@accounts/server` throws an error while compiling.
It's related to `emittery` package: sindresorhus/emittery/issues/34

Instead of asking a user to resolve to `emittery@0.4.0` I think `@accounts/server` should require the correct version that works fine.

Also, could it be released ASAP since it breaks apps with latest typescript?

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,14 +16,19 @@
     "coverage": "lerna run coverage",
     "codecov": "codecov"
   },
-  "workspaces": ["packages/*"],
+  "workspaces": [
+    "packages/*"
+  ],
   "husky": {
     "hooks": {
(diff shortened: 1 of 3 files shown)


TimMikeladze:
This looks good me. I do think that `@accounts/server` should require emittery. @pradel @davidyaha @Aetherall thoughts?

davidyaha:
@kamilkisiela  Thanks!

davidyaha:
Published, let me know if that worked.

kamilkisiela:
It worked, toda :+1:

PR #347, opened 2018-06-10
return tokens from loginWithService

1e1f:
return tokens used by loginWithService so they can be used by the calling function.


pradel:
@1e1f Thank you for the pr!

Issue #362, opened 2018-07-03
Improving the graphql package


pradel:
I was thinking about how to improve the graphql package.
- [x] export a `@authenticated` which will have the same behavior as the current `authenticated` function.
- [x] export the resolver object and let the user merge the resolvers himself (with lodash merge for example) instead of `accountsGraphQL.extendWithResolvers`.
- [x] remove the profile key on the CreateUserInput type
- [x] remove the email key on the User type.

Do you have other other ideas?

Aetherall:
> export a @authenticated directive to replace the current authenticated function.
Maybe we should add it instead of replacing it, the authenticated function could be usefull for resolvers where authentication is needed depending on the parameters or the object

>  export the resolver object and let the user merge the resolvers himself instead
I agree :+1:

pradel:
Maybe we can also remove the `email` key on the `User` type. Right now internally we use an array of emails so it make no sense to get that email key which is a string.

pradel:
By default I think we should remove the profile key on the `CreateUserInput` type. We can add docs to let the user know how to extend the type with a profile

Aetherall:
I agree for both of those points :+1:

TimMikeladze:
Started work on the remaining two tasks here https://github.com/accounts-js/accounts/pull/366

PR #366, opened 2018-07-05
feat: @authenticated directive

TimMikeladze:
Modifies createJSAccountsGraphQL to return typeDefs (renamed from schema), resolvers, and schemaDirectives.

Created AuthenticatedDirective which wraps all graphql objects or fields annotated with @authenticated.

**Todo**

- [x] Update documentation
- [ ] Update example
- [x] Cache authentication check. A request doesn't need to run through the authenticated logic for each type or field annotated with the directive, just once and caching the result should suffice. 

**Questions**

- Should `@authenticated` to be renamed to `@auth` to be more concise?
- Should `createJSAccountsGraphQL` be renamed to `createAccountsGraphQL`?

**Related issues**: #362

### DIFF
--- a/packages/graphql-api/README.md
+++ b/packages/graphql-api/README.md
@@ -49,11 +49,11 @@ const typeDefs = [
   type Query {
     myQuery: String
   }
-  
+
   type Mutation {
     myMutation: String
   }
-  
+
   schema {
       query: Query,
       mutation: Mutation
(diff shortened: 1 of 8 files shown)


Aetherall:
> Should @authenticated to be renamed to @auth to be more concise?
Yes I think it's easier to use  :+1: 

> Should createJSAccountsGraphQL be renamed to createAccountsGraphQL?
Yes :+1:

TimMikeladze:
I think code wise this is ready. Just documentation remaining.

PR #369, opened 2018-07-06
Fix infinite loop issue during token refresh for graphql-client

pradel:
If we are executiong a refresh token mutation do not call refress session again otherwise it will end up in an infinite loop


rankun203:
Good job on this one 👍I didn't seem to really solved this issue.

PR #371, opened 2018-07-09
use fetch for the graphql client

pradel:
New constructor:
```js
const client = new GraphQLClient({
  url: 'http://....',
})
```


pradel:
If we use fetch we will not be able to have ws transport.

PR #384, opened 2018-07-13
Setup e2e testing

pradel:
The goal of this pr is to implement e2e for the suite. I am trying an approach that will allow us to write generic tests that should work no matter which database or transport is used.
I am using the accounts client in order to do the requests.
Any feedback welcome 🙌

### DIFF
--- a/packages/e2e/__tests__/servers/server.ts
+++ b/packages/e2e/__tests__/servers/server.ts
@@ -9,6 +9,7 @@ import { AccountsPassword } from '@accounts/password';
 // TODO rename to AccountsMongo ?
 import { Mongo } from '@accounts/mongo';
 import { createAccountsGraphQL, accountsContext } from '@accounts/graphql-api';
+import { User } from '@accounts/types';
 
 // Client
 import { AccountsClient } from '@accounts/client';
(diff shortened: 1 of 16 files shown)


packages/e2e/__tests__/servers/server.ts
```diff
@@ -0,0 +1,154 @@
+import * as mongoose from 'mongoose';
+import { ApolloServer, gql } from 'apollo-server';
+import { merge } from 'lodash';
+import fetch from 'node-fetch';
+
+// Server
+import { AccountsServer } from '@accounts/server';
+import { AccountsPassword } from '@accounts/password';
+// TODO rename to AccountsMongo ?
+import { Mongo } from '@accounts/mongo';
```
TimMikeladze: This server instance is tied to mongo at the moment. If the database is parameterized then we could test all the database drivers using the same set of tests.

pradel: done

pradel:
Ready for review.
The tests are run against the password module only for now.
Tests for the `rest` and `graphql` server using the mongo database.

pradel:
They will help us catch the regressions that we might introduce in the future.
More tests can be added on separate pull requests.

TimMikeladze:
@pradel Could you please resend this PR? I accidentally merged this branch. There's a reverted commit now on master in order to fix this. Sorry.

PR #402, opened 2018-07-20
Custom error messages

pradel:
I would like to have your opinion on this pr, I want to add custom errors to the different packages. What do you think about this simple approach? @TimMikeladze @davidyaha @Aetherall


pradel:
The problem with that approach is that we will need to have the strings of all services stored inside accounts-server no ?

TimMikeladze:
Each service implementation would provide their own default labels object which would be merged with the corresponding key in `accounts.options.labels`. The services already have a reference to accountsServer passed in, this can be an additional step in that code.

TimMikeladze:
@pradel This PR looks good.

What are your thoughts on the following scenario: in the #487 PR for react components an "invalid password" or "user not found" are displayed as form level errors instead of field level errors because there is no additional metadata included when throwing the error indicating which field the error pertains too. This isn't high priority but something to think about in related to errors.

pradel:
Yeah you are right, don't have a good idea how to handle that for both rest and graphql...
Btw "invalid password" or "user not found" when login (reset-password, etc..) allow user enumeration which for security reasons is not good, I think it should be disabled by default 🙈.

PR #403, opened 2018-07-22
Add html email templates

lorensr:
I think it'll give a better idea to people of the existence of the option to have html if they see it printed to console or if they look at the code.


pradel:
Good idea! can you please fix the `EmailTemplateType` type so the ci will pass? Thanks :)

lorensr:
Right, Typescript! Added the type, but CI still failing:

```
$ prettier --list-different '**/*.{json,md,js,ts}'
packages/server/src/utils/email.ts
error An unexpected error occurred: "Command failed.
Exit code: 1
Command: sh
Arguments: -c prettier --list-different '**/*.{json,md,js,ts}'
Directory: /home/circleci/project
Output:
".
```

When I run the command locally, I get:

image

Issue #412, opened 2018-07-28
Remove parameter from getUser?


lorensr:
`graphql-client.js`:
```
  public async getUser(accessToken: string): Promise<User> {
```

The server should be able to use the access token from the request header / get the user from `context.user`—it would be nicer/simpler if I didn't have to provide the parameter.

pradel:
👍

elie222:
Same would apply to `logout` I take it.

elie222:
Made a PR for the fix (for getUser):
https://github.com/accounts-js/accounts/pull/420

May need to still change something around the tests for it.

TimMikeladze:
Closed by #420 

Published `0.3.0-beta.25`

PR #420, opened 2018-07-30
Made access token optional when calling getUser with graphql

elie222:


pradel:
Maybe for this one it makes sense to just remove the parameter what do you think?

elie222:
Could be. I figured I'd touch it as little as possible and still give
people the previous option if it was needed. But I'm happy to remove it as
well.

Elie

pradel:
If you can remove them would be super cool. The other logged in mutations (eg change password) require you to send the token via headers and do not accept a token parameter. If we remove them it will be more consistent :)

elie222:
Makes sense. Will remove shortly.

elie222:
Removed.

Was also wondering if we need to send `accessToken` as an arg for these two items:

```gql
  impersonate(accessToken: String!, username: String!): ImpersonateReturn
  refreshTokens(accessToken: String!, refreshToken: String!): LoginResult
```

For `impersonate` perhaps make sense to keep? But not for refresh tokens?

pradel:
I guess for now we can leave it like this because if we change it we have to change the rest package and test it. Can be done in another pr maybe?

pradel:
Close #412

TimMikeladze:
Thank you!

Published `0.3.0-beta.26`

PR #423, opened 2018-07-31
Fix to accounts link sending null string as access token

elie222:
`'null'` was being sent to the server (rather than `null`) which was throwing errors of bad access token rather than just skipping the check.

Was sending me annoying errors to the client, so I fixed it.


elie222:
Fixing the failing test now

pradel:
The failing test is not from . your pr you can ignore it I will rerun the ci manually

PR #437, opened 2018-08-08
Load the user in the context resolver

### LINKED ISSUE #421: Automatically adding user to graphql resolver context
I see the code has `authenticated` as shown here:

```
changePassword: authenticated(accountsServer, changePassword(accountsServer)),
```

Would it make sense to just add authenticated to the context and have the `user` object available for every request made to the server?

Right now I have this in my code:

```ts
const server = new ApolloServer({
  typeDefs: [typeDefs as any, accountsGraphQL.typeDefs],
  resolvers: merge(accountsGraphQL.resolvers, resolvers),
  context: async ({ req }) => {
    const ctx = accountsContext(req)
    const { authToken } = ctx

    let user

    try {
      user = await accountsServer.resumeSession(authToken as string)
    } catch (error) {}

    return {
      ...ctx,
      user,
      userId: user && user._id,
    }
  },
})
```

Seems something like this would be useful for all users of the `graphql-api` package. Would it make sense to have `accountsContext` also return the user object if it exists.

pradel:
Close #421 
The new context will now be async (breaking change)

### DIFF
--- a/packages/e2e/__tests__/servers/server-graphql.ts
+++ b/packages/e2e/__tests__/servers/server-graphql.ts
@@ -85,11 +85,10 @@ export class ServerGraphqlTest implements ServerTestInterface {
     this.apolloServer = new ApolloServer({
       typeDefs: [gql(accountsGraphQL.typeDefs), typeDefs],
       resolvers: merge(accountsGraphQL.resolvers),
-      context: async ({ req }: any) => ({
-        ...(await accountsContext(req, {
+      context: ({ req }: any) =>
+        accountsContext(req, {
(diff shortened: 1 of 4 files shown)


lorensr:
```js
      context: async ({ req }: any) => ({
        ...(await accountsContext(req, {
          accountsServer: this.accountsServer,
        })),
```
equivalent to:
```js
      context: ({ req }: any) => accountsContext(req, {
          accountsServer: this.accountsServer,
        })
```

pradel:
Yeah e2e tests are passing so ... :)

PR #469, opened 2018-09-27
more info on where AccountsClient came from

williamli:


williamli:
How do I get accountsTransport ??

```
const accountsTransport = // my accounts-js transport, rest or graphql
const accounts = new AccountsClient({
  // options
}, accountsTransport);
```

PR #476, opened 2018-10-10
[@accounts/password] Support uppercase emails when sending the different emails (verification, rest, ...)

pradel:
Currently you can't reset your password if you are trying to send an email with an uppercase.

### DIFF
--- a/packages/password/__tests__/__snapshots__/accounts-password.ts.snap
+++ b/packages/password/__tests__/__snapshots__/accounts-password.ts.snap
@@ -61,6 +61,8 @@ Array [
 ]
 `;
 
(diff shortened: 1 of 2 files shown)


packages/password/src/accounts-password.ts
```diff
@@ -256,18 +251,14 @@ export default class AccountsPassword implements AuthenticationService {
...
-    if (!address) {
+    if (!address || !isString(address)) {
       throw new Error('Invalid email');
     }
+
     const user = await this.db.findUserByEmail(address);
     if (!user) {
       throw new Error('User not found');
     }
-    // Make sure the address is valid
```
pradel: Here we actually don't need this check since we are calling `this.db.findUserByEmail` to get the user.

PR #479, opened 2018-10-13
Fix breaking change

1e1f:
commit #420 broke my implementation by 'making accessToken optional', in fact the code change there didn't make it optional, it forced accessToken to undefined by ignoring the argument and replacing with an empty object. This change should safely allow for a nullable argument of any tokens to pass through this module.

server versions used for testing
    "@accounts/mongo": "0.3.0-beta.29",
    "@accounts/password": "0.3.0-beta.29",
    "@accounts/rest-express": "0.3.0-beta.29",

### DIFF
--- a/packages/rest-client/src/rest-client.ts
+++ b/packages/rest-client/src/rest-client.ts
@@ -108,10 +108,10 @@ export class RestClient implements TransportInterface {
     return this.authFetch('logout', args, customHeaders);
   }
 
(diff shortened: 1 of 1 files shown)


pradel:
Okay I see the problem, we are extracting the token from `req.body` on the `rest-express` package. Maybe a better fix for this issue would be to extract it from the headers on the server side (also logout must have the same problem I guess).

pradel:
We will have to finish the e2e tests for this part to avoid regressions in the future.

PR #485, opened 2018-10-21
Changelog

TimMikeladze:


pradel:
I like the idea to auto generate the changelog 🎉 
Do you know https://github.com/semantic-release/semantic-release? Maybe it can also be a good fit for this feature?

TimMikeladze:
I looked at semantic release. Seems like Lerna already handles some of the release management stuff for us already.

pradel:
Sounds good :) do you know if we can add the notes to the github releases with this tool?

TimMikeladze:
I'm not sure, I think it just generates the changelog.

PR #486, opened 2018-10-23
Updated graphql examples to latest accounts packages

davidyaha:
Had some issues with implementing the graphql packages into a new project.
Decided to try it out on the examples and had to upgrade the dependencies here.
Everything seems to be working so the issue must be on my side.
Been out for quite a while now so it's good to see that the project is thriving!


TimMikeladze:
Keep in mind the register user resolver was renamed to createUser

PR #487, opened 2018-10-24
[WIP] React and Apollo integration

TimMikeladze:
- `@accounts/apollo` package which streamlines creating the accounts client with graphql
- `@accounts/react` handles the bulk of the logic for user sign in, sign up, and authentication
- `@accounts/react-material-ui` - accounts components based on `@material-ui/core`
- Added a `getAccountsOptions` function which retrieves site specific accounts options such as the services enabled or the site title. This is used by the `@accounts/react` to present the user with the relevant components.


ardatan:
You can use Stencil and Ionic that can make accounts-js ui components platform agnostic. So, the same components can be used for all frameworks such as React, Angular and Vue.

TimMikeladze:
@ardatan Interesting suggestion, I haven't heard of those tools before, can you point me in the right direction on some reading materials or where to get started?

I need to evaluate how much effort this would take...I kind of want to wrap up these react PR this week and move on to other work.

PR #490, opened 2018-10-25
Feature/fix tests types

pradel:
Fix the typescript types inside the tests.

Basically just add a ton of any to make the compiler happy .. 🐞


tsconfig.json
```diff
@@ -8,6 +8,7 @@
     "pretty": true,
     "lib": ["es6", "es2015", "es2016", "es2017", "dom", "esnext.asynciterable"],
     "noUnusedLocals": true,
-    "strict": true
+    "strict": true,
+    "allowSyntheticDefaultImports": true
```
TimMikeladze: Maybe enable `esModuleInterop` as well?

pradel: Okay I will try

pradel: It create more errors to fix, let's merge for now and enable it later :)

PR #497, opened 2018-11-15
Fix GraphQL-Modules version

ardatan:
- Moving @graphql-modules/core to peerDependencies instead of directDependencies.
- Use recent stable version

Example;
https://github.com/ardatan/graphql-modules-accountsjs-boilerplate


packages/graphql-api/package.json
```diff
@@ -34,18 +34,19 @@
...
     "nock": "10.0.1",
-    "ts-node": "7.0.1"
+    "ts-node": "7.0.1",
+    "@graphql-modules/core": "0.2.1",
+    "reflect-metadata": "0.1.12"
   },
   "peerDependencies": {
-    "graphql-tools": "^4.0.3"
+    "graphql-tools": "^4.0.3",
+    "@graphql-modules/core": "^0.2.1",
```
pradel: is it a good idea to make it a peerDependencies if users can use the module without it? 🤔

dotansimha: It's the only way to avoid multiple copies of a specific dependency, but still make sure that users know they need to install it.

pradel: I am not a big fan of forcing users to install it if they don't need it ... Separate minor versions are not compatibles between them?

Also other question, what is `reflect-metadata`?

packages/e2e/__tests__/servers/server-graphql.ts
```diff
@@ -1,3 +1,4 @@
+import 'reflect-metadata';
```
TimMikeladze: What is `reflect-metadata` being used for in this PR?

packages/graphql-api/package.json
ardatan: `reflect-metadata` is required for DI, but now it is optional. Then I removed it from the code and introduction.

packages/e2e/__tests__/servers/server-graphql.ts
ardatan: As I said above, `reflect-metadata` is required for DI, but now it is optional. Then I removed it from the code and introduction :)

TimMikeladze:
Thanks for the PR! Merged and published.

ardatan:
Here is the article about this :)
https://medium.com/the-guild/authentication-with-accountsjs-graphql-modules-e0fb9799a9da

ardatan:
We shared this article on Twitter as well
https://twitter.com/ardatanrikulu/status/1063454854255923200

PR #498, opened 2018-11-15
User enumeration

### LINKED ISSUE #337: Option to obfuscate user infos
Would be great to have an option to obfuscate the user infos. So for example instead of 'user not found' we need to say 'Invalid credentials' on the login error messages (and others)

pradel:
Close #337 

- **Breaking change:** new server options `ambiguousErrorMessages` that default to true.
When this option is true all the messages that allow user enumeration are hidden:
- when trying to login with wrong credentials before was throwing with `User not found`, now `Invalid credentials`
- when trying to send an email to reset the password if the user email is not found fail silently
- when trying to send an email to verify the password if the user email is not found fail silently
- when trying to create a new user, we return null to the client instead of the user id (the transports are intercepting the id and return null) if user already exist the function will fail silently
- when trying to impersonate, we return `{ authorized: false }`
- **Breaking change:** the register route of the `rest-express` package now return the userId as a string instead of `{ userId: ''}`

### DIFF
--- a/packages/e2e/__tests__/password.ts
+++ b/packages/e2e/__tests__/password.ts
@@ -4,7 +4,6 @@ const user = {
   email: 'johnDoe@gmail.com',
   password: 'notSecure',
 };
-let userId: string;
 
 Object.keys(servers).forEach(key => {
   const server = servers[key];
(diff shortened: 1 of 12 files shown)


pradel:
I am not a big fan of the name `ambiguousErrorMessages` so if you have a better idea 🤔

PR #505, opened 2018-11-23
return LoginResult from resetPassword

benawad:
Added this to be able to automatically login the user after reseting their password.

### DIFF
--- a/packages/client-password/__tests__/client-password.ts
+++ b/packages/client-password/__tests__/client-password.ts
@@ -1,10 +1,18 @@
 import { AccountsClientPassword } from '../src/client-password';
 
+const loggedInResponse = {
+  sessionId: '1',
+  tokens: {
+    accessToken: 'accessToken',
+    refreshToken: 'refreshToken',
+  },
+};
+
 const mockedClient = {
   transport: {
     createUser: jest.fn(),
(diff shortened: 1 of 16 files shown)


pradel:
Hey thanks for the pr that's a good idea!
I was wondering, maybe we could have this feature enabled by an accounts-password option (boolean) rather than having it by default, what do you think?

benawad:
sounds good

benawad:
So if we make this opt-in what do you want to return from the mutation?  

A union?
```
union ResetPasswordResponse = LoginResult | Boolean
resetPassword(token: String!, newPassword: String!): ResetPasswordResponse
```

Return null if they don't opt-in?
```
resetPassword(token: String!, newPassword: String!): LoginResult
```

Maybe create a new mutation for it?
```
resetPasswordAndLogin(token: String!, newPassword: String!): LoginResult
```

pradel:
I think I prefer the "Return null if they don't opt-in?" way, what do you think @TimMikeladze?

TimMikeladze:
Agreed. Return null.

benawad:
I made it opt-in, let me know if you would like to see any other changes

TimMikeladze:
@benawad Thank you very much for the PR. 

Published as `0.7.0`

PR #508, opened 2018-11-27
Tests for the graphql-api package

pradel:


packages/graphql-api/__tests__/modules/accounts/index.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+describe('AccountsModule', () => {
+  describe('withSchemaDefinition', () => {
+    it('should export typeDefs with schema definition', () => {
+      const accountsGraphQL = AccountsModule.forRoot({
+        accountsServer: accountsServer as any,
+        withSchemaDefinition: true,
+      });
+      expect(accountsGraphQL.typeDefs).toMatch(/schema/);
```
ardatan: Could you use `print(accountsGraphQL.typeDefs)` here?

pradel: 👍

pradel:
@ardatan I updated the pr with the last version which fixed the `extendTypeDefs` option but `withSchemaDefinition` option is still not working unfortunately

ardatan:
Could you try it with latest version 0.2.16?

pradel:
Looks like it's still failing

ardatan:
Oh sorry, it is my fault. I forgot to include the changes in latest build. So, it should be fixed in 0.2.17.

pradel:
@ardatan thanks it worked 🎉

PR #510, opened 2018-11-28
Show the api docs for all the packages

pradel:
Types are generated by typedoc.
Inspired by graphql-modules for that 🎉


aslafy-z:
You may want to change the path prefix: ```/Users/leo/dev/accounts-js```

As in here: https://github.com/accounts-js/accounts/pull/510/files#diff-691461362a3d6e161754c16706d27cfcR27

TimMikeladze:
@pradel Can you add a link to the API docs to the repo's README?

PR #513, opened 2018-12-01
Update GraphQL Modules and GraphQL Code Generator

ardatan:
GraphQL-Modules and GraphQL Code Generator have some changes in their API after new releases;
So this PR does the necessary changes.


packages/graphql-api/src/modules/accounts/schema/types.ts
```diff
@@ -1,6 +1,7 @@
 import gql from 'graphql-tag';
+import { AccountsModuleConfig } from '..';
 
-export default gql`
+export default ({ userAsInterface }: AccountsModuleConfig) => gql`
```
TimMikeladze: Do you think the usage of setting up the `User` interface should be added to the documentation? A note about this in the readme and graphql docs would be useful I think.

ardatan: Oh sorry, I forgot to add this change in the docs. I can remove this if you want.
Otherwise, I can add this in the docs.

TimMikeladze: I think it's useful, please add to the docs.

ardatan: I added some info to docs. Let me know if it is good.

TimMikeladze:
Thank you very much for the PR. 

Published version `0.9.0`

Issue #368, opened 2018-07-05
Documentation


TimMikeladze:
This issue is to track parts of accounts-js which need to be documented. 

API documentation

- [ ] It would be a good idea to generate api documentation for all the package using the typings. 

Usage documentation: 

- [ ] Getting started
  - [ ] REST
  - [ ] GraphQL
  - [ ] React
- [ ] Cookbook
  - [ ] accounts-js GraphQL micro-service and schema stitching
- [ ] Email
- [ ] Strategies
   - [ ] Passwords 
   - [ ] Two factor
- [ ] Sessions
- [ ] Databases
  - [ ] Mongo

pradel:
> It would be a good idea to generate api documentation for all the package using the typings.

💯 do you know a good lib to parse the typescript JsDoc?

TimMikeladze:
https://github.com/TypeStrong/typedoc

pradel:
So the idea will be to use this tool to generate some json and then convert the json to a markdown file?

pradel:
just found this project https://github.com/Microsoft/tsdoc

rjseymour66:
Hi,

I am new to Github, but I am a professional technical writer and would like to help out with your documentation.  Please let me know if you still need help with the user documentation outlined at the beginning of this thread.

- Ryan

rjseymour66:
Sounds great! I will be able to take a look in the next few days and get back to you with some ideas.

I look forward to working with you!

rjseymour66:
Hey sorry for the delay - I have had a busy few weeks!

I think a good first step would be to get the existing documentation in one place.  The existing documentation is a great start, and putting it in context will help.

I can use MkDocs to create a static site and deploy using Github Pages - what do you think?  Then we can chip away at the to do list.

TimMikeladze:
We've got a documentation site deployed already at https://accounts-js.netlify.com using docusaurus. See the `website` and `docs` folders in the repo's root.

rjseymour66:
Ahh gotcha - my bad - still learning to navigate Github.

How about I start with the ReadMe?  I think it looks really good right now, but there are a few minor grammatical fixes and other minor changes that could help.  Also, it would help me to learn the app.  Does that sound good?  If so, I should be able to start on it this weekend.

TimMikeladze:
[...]

PR #514, opened 2018-12-01
[doc](boost-usage): Replace AccountsServer with AccountsBoost

miller-productions:
Replace AccountsServer with AccountsBoost in usage doc for boost README

contributes to accounts-js/accounts#368

as discussed here https://github.com/accounts-js/accounts/issues/368#issuecomment-443282080


TimMikeladze:
Thank you for the PR.

PR #520, opened 2018-12-04
Use a function for AccountsClient in apollo-link

### LINKED ISSUE #470: [HELP] How to create AccountsClient to generate the authLink required in GraphQL?
In 
https://accounts-js.netlify.com/docs/transport/graphql/#hook-it-up-to-the-apollo-client

```
import { accountsLink } from '@accounts/apollo-link';

const accountsClient = new AccountsClient( ... );
const authLink = accountsLink(accountsClient);

export const apolloClient = new ApolloClient({
  link: ApolloLink.from([authLink, httpLink]),
  cache,
});
```

What is an AccountsClient?

Is it `import { AccountsClient } from '@accounts/client';` ?
If so, what is the _accountsTransport_ required to construct AccountsClient?

ardatan:
Fixes #470 
Using a function will make available to handle circular imports;
```ts
accountsLink(() => accountsClient)
```

### DIFF
--- a/packages/apollo-link-accounts/src/index.ts
+++ b/packages/apollo-link-accounts/src/index.ts
@@ -1,15 +1,14 @@
 import { setContext } from 'apollo-link-context';
 import { ApolloLink } from 'apollo-link';
-import { AccountsClient } from '@accounts/client';
 
-export const accountsLink = (accountsClient: AccountsClient): ApolloLink => {
+export const accountsLink = ({ headerName = 'accounts-access-token' }): ApolloLink => {
   return setContext(async (_, { headers: headersWithoutTokens }) => {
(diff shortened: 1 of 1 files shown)


pradel:
The problem to not use the client in the link is that it will not work with react-native storage or any kind of custom storage that the user can implement.

ardatan:
How about factory function :)

TimMikeladze:
@ardatan  I think a factory function would work. @pradel What are your thoughts?

pradel:
@TimMikeladze @ardatan looks like this can fix our problem for now yes!

We will need to update the docs and examples before releasing a new version with this one :)

pradel:
@ardatan can I close this one?

ardatan:
@pradel Yes!

PR #522, opened 2018-12-04
Try to use user._id or user.id

### LINKED ISSUE #521: Need to change the current graphql User resolvers
On the pull request #513 this new resolver has been introduced:
https://github.com/accounts-js/accounts/commit/e42e6c7ee12c2fe9b828c725c10eac1a49eb6343#diff-7c84961d16582309b79b1a9c44c7f1e3R2
We need to change the resolver to not use `_id` which will only work with mongo and fail with other databases

TimMikeladze:
Closes #521 

@pradel This is a pretty simple fix, do you think is sufficient or should we always ensure that `user.id` is set even if mongo is being used?

### DIFF
--- a/packages/database-mongo/__tests__/database-tests.ts
+++ b/packages/database-mongo/__tests__/database-tests.ts
@@ -27,7 +27,10 @@ export class DatabaseTests {
 
   public createConnection = async () => {
     const url = 'mongodb://localhost:27017';
-    this.client = await mongodb.MongoClient.connect(url);
+    this.client = await mongodb.MongoClient.connect(
+      url,
+      { useNewUrlParser: true }
+    );
     this.db = this.client.db('accounts-mongo-tests');
     this.database = new Mongo(this.db, this.options);
   };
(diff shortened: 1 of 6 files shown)


pradel:
@TimMikeladze actually I think it make more sense to return the id as a string from the mongo database package.
Currently we are doing like this:
```js
public async findUserById(userId: string): Promise<User | null> {
    const id = this.options.convertUserIdToMongoObjectId ? toMongoID(userId) : userId;
    const user = await this.collection.findOne({ _id: id });
    if (user) {
      user.id = user._id;
    }
    return user;
  }
```
But we can change it to:
```
public async findUserById(userId: string): Promise<User | null> {
    const id = this.options.convertUserIdToMongoObjectId ? toMongoID(userId) : userId;
    const user = await this.collection.findOne({ _id: id });
    if (user) {
      user.id = user._id.toString();
    }
    return user;
  }
```

pradel:
In that way we don't need the user resolver

TimMikeladze:
@pradel I understand, I can make these changes.

TimMikeladze:
Thank you for fixing my mistakes and finishing this issue @pradel 👍

PR #525, opened 2018-12-14
Remove ambiguousErrorMessages from `createUser`

TimMikeladze:
The created user's id should always be returned from `createUser` even if ambiguousErrorMessages is set to true.

**todo**

- [ ] update tests

### DIFF
--- a/packages/e2e/__tests__/password.ts
+++ b/packages/e2e/__tests__/password.ts
@@ -18,7 +18,7 @@ Object.keys(servers).forEach(key => {
           email: user.email,
           password: user.password,
         });
-        expect(userId).toBeNull();
+        expect(userId).toBeTruthy();
       });
     });
 
--- a/packages/graphql-api/src/modules/accounts-password/resolvers/mutation.ts
+++ b/packages/graphql-api/src/modules/accounts-password/resolvers/mutation.ts
@@ -16,7 +16,7 @@ export const Mutation: MutationResolvers<ModuleContext<AccountsModuleContext>> =
   },
   createUser: async (_, { user }, { injector }) => {
     const userId = await injector.get(AccountsPassword).createUser(user as PasswordCreateUserType);
-    return injector.get(AccountsServer).options.ambiguousErrorMessages ? null : userId;
+    return userId;
   },
   twoFactorSet: async (_, { code, secret }, { user, injector }) => {
     // Make sure user is logged in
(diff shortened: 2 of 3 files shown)


pradel:
Hey @TimMikeladze why this change ?
If we return an id this open the function to user enumeration.

TimMikeladze:
How does this open up to user enumeration? The only time createUser would be called is either by the new user (in which case they can know their own user ID, that's not a secret) or it would be called as part of your business logic server side, in which case it's useful to get the id of the created user.

I don't think revealing the id of the user to themselves presents any security risks?

pradel:
Currently the server side function always return the id, just the transports are not returning the id.

The createUser function should fail silently if the email we try to create already exist, so no id returned. If I am an attacker and I want to discover which user registered to your service I will call the createUser endpoint with a random email. If I get the id back means the email does not exist in the system, but if the function return null means the user exist.
If we never return the id to the client an attacker will not be able to guess if the user exist or not.

Don't know if it's clear or not 😅

TimMikeladze:
That explanation makes sense.

sakulstra:
@pradel I'm still wondering how this helps with user enumeration :thinking:

1. I create a user via `createUser`
2. I authorize this user via `authorize`

-> I still know if the user already existed or not based on the possibility to login or not(because I think login even works with unverified emails).

Also I tested the current behavior and I think it doesn't work at all.
There's always an `"message": "Email already exists",` error thrown in graphql as the error is directly thrown to the server so `ambiguousErrorMessages` is skipped.

pradel:
@sakulstra 

> -> I still know if the user already existed or not based on the possibility to login or not(because I think login even works with unverified emails).

This is true but you need to know the user password for that ;) 

> There's always an "message": "Email already exists", error thrown in graphql as the error is directly thrown to the server so ambiguousErrorMessages is skipped.

This looks bad indeed, do you mind opening an issue with your findings please?

sakulstra:
> This is true but you need to know the user password for that ;)
I think you don't? Accounts-js doesn't disallow authentication with an unverified email-address, so you can signUp & signIn -> if you cannot signIn the user existed already.

pradel:
> I think you don't? Accounts-js doesn't disallow authentication with an unverified email-address, so you can signUp & signIn -> if you cannot signIn the user existed already.

By default you are right, I still think it should be the default behavior, it's up to you to setup this check when the user is logging in, you can do it using the login hooks.

PR #555, opened 2019-01-18
Added TypeORM database driver

birkir:
[TypeORM](http://typeorm.io) is an ORM with support for MySQL, MariaDB, PostgreSQL, SQLite, MSSQL, Oracle and experimental Mongodb support.

This can also be used as base for Sequelize or similar libraries.

Noticeable changes
 - Now using UUID as id's for tests ([because ints are bad](https://www.clever-cloud.com/blog/engineering/2015/05/20/why-auto-increment-is-a-terrible-idea/))
 - Added postgres image to circleci
 - Services and emails are stored in a separate table for relational consistency.
 - Sessions have hard relationships to Users (with cascading), for testing they can be null, so I had to add errors on accounts-server and appropriate type union.
 - Added e2e support for typeorm but didn't enable it by default as the test doesn't support multiple database drivers.

### DIFF
--- a/packages/types/src/types/database-interface.ts
+++ b/packages/types/src/types/database-interface.ts
@@ -70,7 +70,7 @@ export interface DatabaseInterfaceSessions {
     token: string,
     connection: ConnectionInformations,
     extraData?: object
-  ): Promise<string>;
+  ): Promise<string | null>;
 
   updateSession(sessionId: string, connection: ConnectionInformations): Promise<void>;
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,10 @@ jobs:
       - image: circleci/node:9
       - image: circleci/mongo:3
       - image: redis:4@sha256:a471807cc636a02d8be3959a52dd7262a83ce6351be19605ce17d5da06c42f1d
+      - image: circleci/postgres:10.6
+        environment:
+          POSTGRES_USER: accounts-js
+          POSTGRES_DB: accounts-js-tests-e2e
     steps:
       - checkout
       - run: yarn install --frozen-lockfile
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "opencollective": "1.0.3",
     "prettier": "1.15.3",
     "ts-jest": "23.10.5",
-    "tslint": "5.12.0",
+    "tslint": "5.12.1",
     "typedoc": "0.13.0",
     "typedoc-plugin-docusaurus": "1.0.14",
     "typescript": "3.2.2"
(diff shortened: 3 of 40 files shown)


birkir:
Codecoverage doesn't seem to recognize decorators.

Also I am not sure how to actually test "them" specifically

pradel:
Hey, @birkir thanks for the pr! I will try to find some time to review it soon.

Do you think you can take a look to setup the end to end tests for it?
You can take a look at the [mongo tests](https://github.com/accounts-js/accounts/blob/master/packages/database-mongo/__tests__/database-tests.ts) or [redis tests](https://github.com/accounts-js/accounts/blob/master/packages/database-redis/__tests__/database-tests.ts) for the database only tests.

We also have another end to end test suite for now only setup for mongo, you can check the code there: https://github.com/accounts-js/accounts/blob/master/packages/e2e/__tests__/databases/mongo.ts

Happy to help if you have any question :)

TimMikeladze:
This is fantastic! Code looks good. Getting some tests, additional documentation and maybe an example around this would be great. We should really encourage typeorm as the default database driver.

Also adding an entry into the `databasePackages` for `@accounts/typeorm` within `@accounts/boost` package would be nice too for streamlining the accounts graphql server creation.

birkir:
Great. I'll work on it this weekend! Thanks for the feedback

birkir:
Got a little jump start on this. All lights green.

Requesting feedback and help on docs and implementation of:
> We should really encourage typeorm as the default database driver in our docs since it supports all the major databases.

I personally would like to see this get merged first (before promoting) and maybe get some real world usage, bugs or discussion.

Thanks, have a good weekend.

TimMikeladze:
@birkir Sounds good. I can write some additional docs and examples, we certainly don't need them to merge this.

Thanks again for the PR, I'll be merging and publishing a new release shortly.

TimMikeladze:
Published 0.10.0

packages/types/src/types/database-interface.ts
```diff
@@ -70,7 +70,7 @@ export interface DatabaseInterfaceSessions {
     token: string,
     connection: ConnectionInformations,
     extraData?: object
-  ): Promise<string>;
+  ): Promise<string | null>;
```
pradel: @birkir why did you add null there?

birkir: The Session can be null from the database getter methods. I tried to make it as flexible as possible but I really wanted to keep the constrain on User<>Session. So when TypeORM is used for the e2e test, we cannot guarantee that the result from `findSessionById` and `findSessionByToken` returns the ID.

User delete cascades related sessions.

My first design actually removed sessions as they were invalidated, (instead of soft delete), because I didn't really understand the need. I ended up reverting that for compatibility with redis and mongo.

pradel: I see, removing session when invalidating them is a good idea, would be great to create an option for that :)

But this is for a create function, is it really possible that a create function return `null`?

PR #565, opened 2019-01-27
fix: ignore context when no req available

birkir:
If you have other graphql-modules with subscriptions (through websocket), there will be no `req` object and it will fail unrelated subscriptions because of this.

### DIFF
--- a/packages/graphql-api/src/utils/context-builder.ts
+++ b/packages/graphql-api/src/utils/context-builder.ts
@@ -9,6 +9,7 @@ export const context = (moduleName: string) => async (
   _: any,
   { injector }: ModuleSessionInfo
 ) => {
+  if (!req) { return; }
   const config: AccountsModuleConfig = injector.get(ModuleConfig(moduleName));
   const headerName = config.headerName || 'accounts-access-token';
   const authToken = (req.headers[headerName] || req.headers[headerName.toLowerCase()]) as string;


packages/graphql-api/src/utils/context-builder.ts
```diff
@@ -9,6 +9,7 @@ export const context = (moduleName: string) => async (
   _: any,
   { injector }: ModuleSessionInfo
 ) => {
+  if (!req) { return; }
```
ardatan: You should return null here not to break typings

birkir: Fixed!

ardatan: Hmm, it didn't work either

ardatan: Oh wait, we should return `{}` and make all fields of context interface optional.

birkir:
Hmm ok, I have to rethink this. I can fix this for now, but maybe It would be a good idea to actually support subscriptions 👍 

https://github.com/accounts-js/accounts/blob/e42e6c7ee12c2fe9b828c725c10eac1a49eb6343/packages/graphql-api/src/modules/accounts/index.ts#L32-L38

ardatan:
I think it is better to make it like
```
export interface AccountsModuleContext<IUser = User> { 
   authToken?: string; 
   userAgent?: string; 
   ip?: string;
   user?: IUser; 
   userId?: string; 
 } 
```

TimMikeladze:
@birkir Can you get the code coverage test to pass before I merge this?

birkir:
You got it

PR #597, opened 2019-02-19
 [graphql] Update example and docs to show how to extend the user

pradel:


docs/transports/graphql.md
```diff
@@ -204,8 +206,30 @@ input CreateUserProfileInput {
 }
 ```
 
+By default accounts-js only allow 'username', 'email' and 'password' for the user. In order to add custom fields you need to pass the `validateNewUser` function when you instanciate the 'accounts-password' package.
```
TimMikeladze: Typo, `instanciate` should be `instantiate`

Issue #614, opened 2019-03-14
Schema modification to change unsupported column type


kfrajtak:
Hi, 

I am trying to use Accounts graphql module with MySql database in my app and I'm getting

>  Data type "jsonb" in "User.profile" is not supported by "mysql" database.

Is there a way how change the column type for `User.profile` without actually modifying the compiled file (`node_modules\@accounts\typeorm\lib\entity\UserService.js`)? There are more occurrences of the `jsonb` type. 

Thank you,
Karel

birkir:
Yes but my plan was to add non-json column type sql support.

Actually I made sure to already split up the columns so that everything was prepared for this.

Just matter of using JSON.parse/stringify if column type is not json/json.

I will have time for this very soon.

pradel:
@birkir the profile should be managed by the users see #617

lihaibh:
@birkir @pradel there is another jsonb column in UserService class, the options property.
Is it possible to modify it as well?

this type is not compatible with all databases

pradel:
@lihaibh you are right, my bad I will reopen the issue

pradel:
@birkir I don't really know how typeorm is working but is there a way to do some database specific logic? (eg use jsonb for postgres and use another table for mysql?) The problem I can see with json stringify / parse is that it will not be easy for the user to run some queries on these fields.

lihaibh:
@pradel the only workaround I see is to use actual types as entities and OneToOne annotation and if its necessary to add fields we will have to extend these.
so for example for UserService.options:
```
  @Column('jsonb', { nullable: true })
  public options: { bcrypt: string } | any;
```
have to replace with:
```
  @Column({ nullable: true })
  @OneToOne(...)
  public options: UserServiceOptions;
```

Issue #394, opened 2018-07-18
Add additional fields to user model


rankun203:
What is the suggested way to add additional fields to user?

rankun203:
Currently I'm creating a local `UserMeta` just for storing user properties like roles, but still, I'm looking for a better way to do this job.

pradel:
Currently we are providing a profile key to let to the user the ability to add more fields.
I am not super happy with the current solution which is quite limiting.
In my opinion we should remove the profile key and save the whole user object in the database. In that way the user can still continue to use the profile key if he wants to or use the `user` root fields. The user will be able to provide a function to validate the user and only allow some fields for example.

pradel:
Let's use this issue to discuss about this feature :)

TimMikeladze:
Your solution sounds good to me @pradel

pradel:
See https://github.com/accounts-js/accounts/pull/438 which should solve this problem

PR #617, opened 2019-03-17
Remove user profile

### LINKED ISSUE #616: Unable to configure accountsjs with typeorm
I receive an error saying:
Data type "jsonb" in "UserService.options" is not supported by "sqlite" database

In the class UserService there is a field that is a json type and my database is sqlite.
is it possible somehow to make it work with sqlite database?

pradel:
We now have the option to extend the user object, so it's time to remove the profile key on the user.
See https://github.com/accounts-js/accounts/issues/394

Close #616 #614

### DIFF
--- a/examples/graphql-server-typescript/package.json
+++ b/examples/graphql-server-typescript/package.json
@@ -13,12 +13,12 @@
     "test": "yarn build"
   },
   "dependencies": {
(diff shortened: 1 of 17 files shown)


gezquinndesign:
With this change, is there any way to access user.profile.firstName for the email templates after extending? I can't see a way because extension/merging happens after setting up the accountsServer.

PR #644, opened 2019-04-08
[@accounts-password] Remove the unused minimumPasswordLength option and improve the options description

pradel:
Looks like `minimumPasswordLength` is not used


TimMikeladze:
Is there a default function thats used for validating passwords which has the minimum length?

pradel:
@TimMikeladze there is one but it just check that the string is not empty (since the password is hashed on the client it's hard to check for the length on the server)

Issue #635, opened 2019-04-04
Entering an email as a username throws an error


elie222:
The error message shown when entering an email for a username is that a username or email is required. Made it hard for us to debug as a username had been submitted.

derblitz:
I can confirm that the following DOES NOT work:

mutation Register {
  createUser(user: { email: "a@b.com", password: "1234567", profile: {firstName: "a", lastName: "b"} })
}


However, this DOES work:

mutation CreateUser {
  createUser(user: { email: "a@b.com", password: "1234567", profile: {firstName: "a", lastName: "b"} })
}

pradel:
The createUser function will throw if the username does not pass the [default username validation function](https://github.com/accounts-js/accounts/blob/master/packages/password/src/accounts-password.ts#L77), the error message should be improved 👍

elie222:
Thanks. Made a PR:
https://github.com/accounts-js/accounts/pull/646

Still needs some work.

PR #646, opened 2019-04-08
Improve error messages

elie222:
Started making this update in the browser, but I see I need to edit more files.

I still need to add a field here:
https://github.com/accounts-js/accounts/blob/master/packages/password/src/errors.ts

`invalidUsername: 'Invalid username',`

And check/fix tests.

Fixes: https://github.com/accounts-js/accounts/issues/635

Fixes: #647


Ethaan:
Do you think that, instead of `throwing` an error.

something like this, make sense?

```
type Error {
 path: String!
 message: String
}

type Response {
 ok: Boolean
 errors: [Error]
}
```
so we can do something like.

```
    const { ok } = await accountsPassword.login({
      password,
      user: {
        email,
      },
    });
```

elie222:
If there’s an error it will throw an error in any case, no?

pradel:
@Ethaan to get the error you can do like that:
```js
handleSubmit: async ({ email, password }, { setErrors }) => {
  try {
    const response = await accountsPassword.login({
      password,
      user: {
        email,
      },
    });
  } catch (error) {
      setError(error)
    }
},
```

Ethaan:
Thanks @pradel!

pradel:
Thanks for the pr @elie222 I will include it in the next release :)

Issue #607, opened 2019-03-06
accountsBoost overwrites user-provided options


moltar:
If I supply my own `db` via `accountsBoost`, it will always get overwritten:

https://github.com/accounts-js/accounts/blob/4e4dc148c41957ddf8021d0b65c476bed668ffa5/packages/boost/src/index.ts#L67

This basically locks `accountsBoost` only to mongo implementation, and will block `@accounts/typeorm`, which is very desirable!

pradel:
You are right this part should be updated to also support typeorm.
Pull request welcome 🙌

agustif:
> You are right this part should be updated to also support typeorm.
> Pull request welcome 🙌

I would love to try to get this done, but would also highly appreciate any guidance on where should I start.

I myself am planning to use AccountsJS with a TypeORM/PostgreSQL project, and was planning to copy the implementation from @brikir/primecms which I got to know about AccountsJS to start from...

agustif:
Well. I don't know about accountsBoost, but I'm gonna go ahead and try to modify one of the examples to use postgresql instead of the default mongo and try to document submit the message as a PR back once it's finished.

Im going to basically follow onto the lines of @birkir on his implementation of this on primecms, which I recently saw made him develop accounts-js/typeorm package which should allow to do this outside of accountsBoost pretty easy than from start!

Hope to be able to do it by myself, will report back soon!

agustif:
I'm a little bit stuck, since I don't know how to create the postgres database (I do create the postgres service with the docker file, might share my progress so far later today either if I succeed or not

agustif:
OMG I don't even believe I got it working by myself, It really needs a lot of polish, and @birkir really deserves all the credit for creating the @accounts/typeorm package.

But it's working!

What would be the best way to commit this back to the community, It really needs some work to become a mergeable PR, but maybe I can just send it as a draft PR on a new branch called examples/graphql-server-typeorm-postgres and recieve some feedback as comments.

Btw I'm also trying to create a new UI version in antd, and would like to end up adding NextJS to the mix, might commit that later down the road!

I'm just so happy I could get it working!! 😄 
<img width="1680" alt="accounts-typeorm-postgres" src="https://user-images.githubusercontent.com/6601142/56158033-3f7d1680-5fc1-11e9-91a3-2d24b0b80753.png">

birkir:
Good to hear @agustif 

You can start by sharing a PR or branch with your changes. We can give you a hand if necessary or feedback.

[...]

PR #654, opened 2019-04-15
Adding Postgres/Typeorm Server Example and AntDesign Client Example.

agustif:
This is barely working on the server side, and definetely not working on the client side, any help or direction on how to follow, would be greatly appreciated, I don't know how to tackle #607 accountsBoost decoupling of mongo, but would love to help get it working too!

Anyway, thanks for this awesome library, and all the open-source work!


agustif:
Why deleting yarn.lock fucked up the tests?

I deleted it because it had like 10.000 insertions or something like that

agustif:
I might try to split this off into multiple branches/PR's

- typeorm-postgresql
- react-ant.design
- AccountsJS as a GraphQL-Module + PostgreSQL

I hope they help anyone like me trying to figure out from other examples done in Mongo how to make AccountsJS both work with postgreSQL thanks to the amazing typeorm plugin, and also be used as an imported module with graphql-modules.

Anyway, hope splitting into smaller-problems/examples helps it getting to master.

pradel:
@agustif if you can split them into multiple pr's that would be great :)
I am just wondering if we really need the "react-ant.design" pr. It should be exactly the same as the material ui example no?

agustif:
> @agustif if you can split them into multiple pr's that would be great :)
> I am just wondering if we really need the "react-ant.design" pr. It should be exactly the same as the material ui example no?
Hi Leo @pradel, thanks for the feedback! 

Well maybe we can skip that one, I'm just not a big material-ui fan, just started using it only because of it's inclusion in Accounts react example.

Maybe it shouldn't be that hard to reimplement, but it isn't at least a one-to-one implementation of just replacing Buttons and other material-ui tags for ant-design's equivalents, since they have very different API's etc, believe me it was what I first tried.

But probably  no need for it on the main repo!

Thanks again, I will push it later after cleaning up some unneeded stuff regarding ssl and timezone implementation of PRime which I copied it mostly from!

agustif:
We can close this one down, I opened a new one only with the postgres server example: https://github.com/accounts-js/accounts/pull/689

I might prioritize the email-passwordless login over the ant.design implementation, I'm growing fond of Material-UI anyway and now that v4 has shipped I feel compelled to give it a try!

I might open a new example to use both Ant.Design and NextJS, that might be a better showcase of something else or a little more difficult to achieve than just re-doing it in react with another design lib.

PR #697, opened 2019-06-03
Adding TypeORM to docs

### LINKED ISSUE #682: Postgres example/docs
I made a shitty PR on #654 

I re-worked it separating only the example concerning the postgres/graphql/typeorm server.

Here's the repo link.
https://github.com/agustif/typeorm-server-typescript

I added necessary dependencies, @birkir did all the hard work building all the pieces necessary for this to work, and also open sourcing a great example of its use in Prime.

I just used that, and tried to keep it simple stupid.

Added a dockerfile for postgres, even if I just my mac's brew installed postgres instance, anyway...

Cleaning it up a bit further, updating and removing unnecessary dependencies.

Any kind of feedback would be appreciated, I could re-attempt a better PR after it if you think it would add value to the project.

Thanks for creating AccountsJS, awesome stuff!

agustif:
I added a typeorm.md file copying from the mongo one.

Also listed both accounts-boost and the typeorm-postgres example on the examples/README.md

Fixed a typo on accounts-boost README with a double "two two" word. So meta, two times two. Hahaha

Close #682

### DIFF
--- a/docs/databases/typeorm.md
+++ b/docs/databases/typeorm.md
@@ -75,7 +75,7 @@ export const connect = (url = 'postgres://localhost:5432') => {
 ## Example
 
 You can find an example using the database-typeeorm package with a PostgreSQL database in the examples folder
-[graphql-server-typeorm-postgres](https://github.com/accounts-js/accounts/tree/master/examples/graphql-server-typeorm-postgres) 
+[graphql-server-typeorm-postgres](https://github.com/accounts-js/accounts/tree/master/examples/graphql-server-typeorm-postgres)
 
 ## Notice
 
--- a/docs/databases/typeorm.md
+++ b/docs/databases/typeorm.md
@@ -1,0 +1,83 @@
+---
+id: typeorm
+title: Typeorm
+sidebar_label: Typeorm
+---
+
+TypeORM data store for accounts-js
+
+[Github](https://github.com/accounts-js/accounts/tree/master/packages/database-mongo) |
+[npm](https://www.npmjs.com/package/@accounts/typeorm)
+
+## Install
+
+```javascript
+yarn add @accounts/typeorm typeorm pg
+```
+
+## Usage
+
+```typescript
+import AccountsServer from '@accounts/server';
+import typeorm from '@accounts/typeorm';
+import { createConnection } from 'typeorm';
+export const createAccounts = async () => {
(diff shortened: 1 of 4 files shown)


agustif:
Yay! should it be TypeORM for the Sidebar/Title Lable vs the current Typeorm?
Screenshot 2019-06-03 at 16 22 20

agustif:
Yeah I think it does look better.
Screenshot 2019-06-03 at 16 32 50

pradel:
In their docs they write it TypeORM so let's keep it like this 👍

docs/databases/typeorm.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+## Example
+
+You can find an example using the database-typeeorm package with a PostgreSQL database in the examples folder
+[graphql-server-typeorm-postgres](https://github.com/accounts-js/accounts/tree/master/examples/graphql-server-typeorm-postgres)
+
+## Notice
+
+It is our long-term goal for typeorm package to be database agnostic and support all database types it does by default.
```
pradel: The notice is a really good idea but maybe it's better to add it to the top before the install part what do you think?

pradel: maybe let's use the markdown `>` so it's really obvious that users have to read it before using the package :)

docs/databases/typeorm.md
```diff
@@ -0,0 +1,83 @@
...
+yarn add @accounts/typeorm typeorm pg
+```
+
+## Usage
+
+```typescript
+import AccountsServer from '@accounts/server';
+import typeorm from '@accounts/typeorm';
+import { createConnection } from 'typeorm';
+export const createAccounts = async () => {
```
pradel: let's add an empty line to make it more readable :)

agustif: I'm not sure if it was supposed to be '## > Notice' that's what I pushed or just '> Notice'!

but done!

pradel: Sorry it wasn't clear so I pushed a fix :)

pradel:
Thank you @agustif I will merge once the pipelines are green :)

pradel:
Screenshot 2019-06-04 at 17 29 45
 like this we have a nice notice that people can't miss

PR #707, opened 2019-06-16
Fix required properties of AccountsServerOptions

NickBolles:
Tiny fix to require the `db` property, but make `tokenSecret` optional.

https://github.com/accounts-js/accounts/blob/master/packages/server/src/accounts-server.ts#L50-L58


NickBolles:
Eh, this needs a little more work than just changing the parameter, the usages need to be updated. I'm going to close this for now.

PR #708, opened 2019-06-16
Allow GET request for /user & Handle path option being a slash

### LINKED ISSUE #705: Rest/Express (get) /user is a POST handler
I didn't want to open a pull request without confirmation that this is a mistake and not intended behaviour. 

Currently the express package registers the get current user endpoint /user as a POST only handler, what is the meaning behind this? I couldn't get the demo react application without editing this within the node_modules.

See file: https://github.com/accounts-js/accounts/blob/master/packages/rest-express/src/express-middleware.ts#L32

Thanks for the library, hopefully I can contribute back once I'm up to speed.

stevokk:
Two small changes with accompanying tests committed.

1. Add GET handler for /user - In Express there is not simple way to handle two methods calling the same route, it's either all or 1, so I went with duping the route entirely for clarity and future flexibility.

2. If the passed path value is a single forward slash, the express route becomes "//user" (and the same for all routes), this is invalid and protected against.

Both changes have tests.

Let me know if you need anything else from me before merging, I will also be committing a separate pull request for fixes in the front-end React demo.

Close #705


pradel:
Thanks @stevokk I will publish a new version soon!

PR #709, opened 2019-06-16
Fix react rest demo

stevokk:
Small changes:

1. Misspelling of First

2. Incorrectly passing firstName as lastName to the server

3. Home.tsx was looking for the User object at the wrong level

4. App wouldn't run because of a change to the typings file after version 2.0.15 dropping templates, adding this version the devDependencies fixed my problem


pradel:
Thanks for the pr @stevokk looks good !

Issue #748, opened 2019-07-28
Automatic log out after a certain amount of time


elie222:
In a project of ours users are being logged out after a certain amount of time (maybe 30 minutes).

This is actually okay as the client wants it, but I need to hook into that log out event and not finding it in the code. Any pointers? Is there an auto log out mechanism built in or is this happening some other way I'm not aware of?

pradel:
You are right the default session time is 30 minutes, but you can overwrite it to make it work for your needs, for example :
```js
const accountsServer = new AccountsServer({
	# ... your config
	tokenConfigs: {
      accessToken: {
        expiresIn: '10m',
      },
      refreshToken: {
        expiresIn: '5d',
      },
    },
})
```

elie222:
Is there a way to hook into the timeout event, to refresh the page at that time? Or I should just handle that myself?

pradel:
For now you can get the tokens from the client and check when the token will expire manually like this https://github.com/accounts-js/accounts/blob/255b6cf180bd029c13f4a9893c99398b6bca8a7e/packages/client/src/utils.ts#L8-L12

elie222:
> You are right the default session time is 30 minutes, but you can overwrite it to make it work for your needs, for example :

Turns out the default is 90 minutes:
https://github.com/accounts-js/accounts/blob/master/packages/server/src/accounts-server.ts#L30

I'm a little confused about the refresh token though. If set for 7 days, will the client not automatically use that and refresh the access token itself?

pradel:
Hey yes you are right sorry it's 90 minutes.
So the `refreshSession` method on the client should refresh the tokens automatically.
https://github.com/accounts-js/accounts/blob/255b6cf180bd029c13f4a9893c99398b6bca8a7e/packages/client/src/accounts-client.ts#L105-L131

Are you using graphql or rest to query your backend?
How do you inject the tokens in the headers?

pradel:
Are you using `apollo-link-accounts` to inject the token?

elie222:
> Are you using graphql or rest to query your backend?
GraphQL

> How do you inject the tokens in the headers?
Regular AccountsJS (we don't touch headers ourselves). See below.

> Are you using apollo-link-accounts to inject the token?
No

This is the file with all our Apollo Client and AccountsJS code:

```ts
import { ApolloClient } from 'apollo-client'
import { ApolloLink } from 'apollo-link'
import { createUploadLink } from 'apollo-upload-client'
[...]

PR #756, opened 2019-07-31
Feature/link refresh token

elie222:
Fixes https://github.com/accounts-js/accounts/issues/748#issuecomment-516824201


pradel:
@TimMikeladze what do you think about this change?

TimMikeladze:
This looks good.

PR #774, opened 2019-08-20
[discuss] accounts-server token creator

ozsay:
Hi,

In our project, we use a 3rd party authenticator service. This service manages access tokens itself. We wrote an `AuthenticationService` to support it, but it has duplicate code with `accounts-server` (and specific graphql endpoints) since we can't use its `loginWithService` and `refreshTokens` functionality.

The reason why we can't use `loginWithService` is the way we deliver our 3rd party token to the client. The only way for us to do that with `accountsjs` is to use it as a session token. then we create access token and refresh token using `accounts` and deliver them. since `accounts-server` generates the session token itself, `loginWithService` is not capable of supporting this scenario.

The way we thought of solving it, is to provide `accounts-server` with a custom `token creator` which will request the access token from the authenticator service and return it to be used as a session token.

The other problem is with `refreshTokens`. When the authenticator service tells us to refresh its tokens, we currently invalidate the current session and create a new one with the new access token. we want to be able to use `accountsServer` functionality to do that. that's why we want to add `createSessionTokenOnRefresh` option to create a new session token when refreshing tokens.

These two changes allow us to use more of accounts-server and client functionality without implementing it ourselves. 

@davidyaha @elie222 FYI

### DIFF
--- a/packages/server/src/types/accounts-server-options.ts
+++ b/packages/server/src/types/accounts-server-options.ts
@@ -26,5 +26,8 @@ export interface AccountsServerOptions {
   prepareMail?: PrepareMailFunction;
   sendMail?: SendMailType;
   tokenCreator?: TokenCreator;
-  createSessionTokenOnRefresh?: boolean;
+  /**
+   * Creates a new session token each time a user refreshes his access token
+   */
+  createNewSessionTokenOnRefresh?: boolean;
 }
--- a/packages/server/src/types/accounts-server-options.ts
+++ b/packages/server/src/types/accounts-server-options.ts
@@ -5,6 +5,7 @@ import { UserObjectSanitizerFunction } from './user-object-sanitizer-function';
 import { ResumeSessionValidator } from './resume-session-validator';
 import { PrepareMailFunction } from './prepare-mail-function';
 import { SendMailType } from './send-mail-type';
+import { TokenCreator } from './token-creator';
 
 export interface AccountsServerOptions {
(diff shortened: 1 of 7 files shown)


packages/server/src/types/accounts-server-options.ts
```diff
@@ -24,4 +25,6 @@ export interface AccountsServerOptions {
   siteUrl?: string;
   prepareMail?: PrepareMailFunction;
   sendMail?: SendMailType;
+  tokenCreator?: TokenCreator;
```
pradel: Does it needs to be a separate interface or can it be just a function?

packages/server/src/types/accounts-server-options.ts
```diff
@@ -24,4 +25,6 @@ export interface AccountsServerOptions {
...
+  createSessionTokenOnRefresh?: boolean;
```
pradel: Can you add a comment to explain this option please?

pradel: Would the name `createNewSessionTokenOnRefresh` be better? (not so sure what do you think?)

packages/server/src/types/accounts-server-options.ts
ozsay: as I think about our use-case, the creator should receive "static" options. so I think it will be better to implement it as an interface that you can instantiate ratter than passing the options in the function

PR #775, opened 2019-08-20
Graphql api context building changes

ozsay:
Hi,

The current `graphql-api` context builder ([here](https://github.com/accounts-js/accounts/blob/master/packages/graphql-api/src/utils/context-builder.ts)) can only support graphql on regular `apollo-server` (that uses regular http server/express).

When we try to load `accounts-server` on `apollo-server-lamda`, the first parameter (session) is a different object than what is currently used in this package.

In this PR i changed the context factory to accept a requestExtractor function to extract info from the session.

@davidyaha @elie222 FYI


ozsay:
i actually want to redesign it to make it more developer-friendly if possible. currently, it's alot of boilerplate to create a new module (for another authentication service).

I wonder if I should introduce breaking changes here, or create a new package

davidyaha:
What would be a breaking change?

ozsay:
best case scenario, to create the `graphql` endpoint would be:
```js
const graphqlModule = createAccountsGraphqlModule(accountsServer, requestExtractor, [
  {
    service: accountsPassword,
    typeDefs: accountsPasswordTypeDefs,
    resolvers: accountsPasswordResolvers
  },
  {
    service: accountsCode,
    typeDefs: accountsCodeTypeDefs,
    resolvers: accountsCodeResolvers
  }
]);

const apolloServer = new ApolloServer({
    typeDefs: graphqlModule.typeDefs,
    resolvers: graphqlModule.resolvers,
    context: graphqlModule.context,
  });
```

and for that, I need to change the graphql module creation and resolvers...

gezquinndesign:
Hello all. I was hoping to use Netlify functions to host my GraphQL API as a lambda but ran into issues to do with the context object not being set.

It would be great to see this PR merged into the next release along with an example of how to use it with apollo-server-lambda. Would this be possible?

darkbasic:
If this is still an issue please open a new PR for Accounts 1.0.

PR #776, opened 2019-08-21
[discuss] code authentication service

ozsay:
Hi,

I implemented a new authentication service which validates a previously sent code to the client.

Common use-cases:

- login with code provided by sms
- login with code provided by email

The service both sends the code ("prepare authentication") using a code-provider and validates it after the client logs-in.

TODO:

- [x] implement core service
- [x] implement twilio sms provider
- [ ] add unit tests

@davidyaha Thoughts?

### DIFF
--- a/packages/code-provider-sms/src/code-provider-sms.ts
+++ b/packages/code-provider-sms/src/code-provider-sms.ts
@@ -1,0 +1,50 @@
+import { Twilio } from 'twilio';
+
+import { CodeProvider } from '@accounts/code';
+
+export type MessageCreator = (code: string) => string;
+
+export interface TwilioSmsCodeProviderOptions {
+  sid: string;
+  secret: string;
+  phoneNumber?: string;
+  messagingServiceSid?: string;
+  messageCreator?: MessageCreator;
+}
+
+export default class TwilioSmsCodeProvider implements CodeProvider {
+  private messageCreator: MessageCreator;
+  private phoneNumber?: string;
+  private messagingServiceSid?: string;
+  private twilio: Twilio;
+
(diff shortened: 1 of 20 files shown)


packages/code-provider-sms/src/code-provider-sms.ts
```diff
@@ -0,0 +1,50 @@
+import { Twilio } from 'twilio';
```
davidyaha: Although it's safe to say that twilio pretty much dominates this field, I wonder if we should allow the user to supply some SMSService rather then depending on twilio.
Another option is to just call this package code-provider-twilio so that it will serve as an example for others that need different SMS providers. Once we get another feature request about some other provider we could refactor for generic SMS providers.
@pradel thoughts?

ozsay: renaming it to `twilio` will prevent another layer of abstraction that we might not need.. so i'm +1 for doing it.

davidyaha: Wow didn't know about notifme.. it looks great! What do you think @ozsay?

ozsay: it may be useful, though I can't see all the features that `twilio` provides (like `messagingServiceSid`)

pradel:
@ozsay From my understanding of the pr, this service can't be used alone since it does not take care of the user account creation right?

ozsay:
@pradel true. we use it with another service.

pradel:
I guess this should be clarified somewhere in the documentation as we want the service to be able to register a user I think.

PR #777, opened 2019-08-21
[discuss] multi-step-authentication

ozsay:
we have a 2 step login scenario that is not supported currently by accounts:

1. user logs-in with username and password.
2. after successful password authentication, we send an SMS with code to the user to be authenticated with (other PR to support login with code).

only after these 2 steps a user is logged in.

This Authentication strategy supports this (and more) scenario.

The Authenticator accepts 2 or more other Authentication services which are being authenticated one after another. finally, after all authentication steps are successful, a token is created to perform the actual login.

- [x] implement service
- [x] add unit tests
- [ ] API docs

@davidyaha FYI

### DIFF
--- a/packages/multi-step/src/accounts-multi-step.ts
+++ b/packages/multi-step/src/accounts-multi-step.ts
@@ -1,0 +1,114 @@
+import { AuthenticationService, DatabaseInterface, User } from '@accounts/types';
+import { AccountsServer } from '@accounts/server';
+
+import { createServiceId, createToken, validateToken } from './utils/crypto';
+import { StepParams, StepResult, MultiStepLoginParams } from './types';
+
+interface DBService {
+  id: string;
+  nextStep?: number;
+  hashedToken?: string;
+}
+
+type AuthenticationServiceArray = {
+  0: AuthenticationService;
+  1: AuthenticationService;
+} & AuthenticationService[];
+
+export default class MultiStepAuthenticationService implements AuthenticationService {
+  public serviceName = 'multi-step-authentication';
+  public server!: AccountsServer;
+  private db!: DatabaseInterface;
+  private steps: AuthenticationServiceArray;
+
+  constructor(steps: AuthenticationServiceArray) {
+    if (steps.length < 2) {
+      throw new Error();
+    }
+
+    this.steps = steps;
+  }
+
+  public setStore(store: DatabaseInterface) {
+    this.db = store;
+  }
+
+  public async authenticateStep({ index, serviceId, ...params }: StepParams): Promise<StepResult> {
+    if (index === 0) {
+      const user = await this.steps[0].authenticate(params);
+
+      if (!user) {
+        throw new Error();
+      }
+
+      const newServiceId = createServiceId(user);
+
+      await this.db.setService(user.id, this.serviceName, { id: newServiceId, nextStep: 1 });
+
+      return { nextStep: 1, serviceId: newServiceId };
+    } else if (!serviceId) {
+      throw new Error();
+    }
+
+    const userFromServiceId = await this.db.findUserByServiceId(this.serviceName, serviceId);
+
+    if (
+      !userFromServiceId ||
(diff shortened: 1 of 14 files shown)


packages/multi-step/src/accounts-multi-step.ts
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    this.steps = steps;
+  }
+
+  public setStore(store: DatabaseInterface) {
+    this.db = store;
+  }
+
+  public async authenticateStep({ index, serviceId, ...params }: StepParams): Promise<StepResult> {
```
davidyaha: @ozsay Not sure I understand how this works. Could you write a small flow summary in a comment?
Who calls `authenticateStep`?

packages/multi-step/src/utils/crypto.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export function createServiceId(user: User): string {
+  const hash = crypto.createHash('sha256');
+
+  hash.update(user.id);
+
+  return hash.digest('hex');
+}
+
+export function createToken(): { token: string; hash: string } {
```
davidyaha: @ozsay Is that the token we send back to the client?

ozsay: yes

packages/multi-step/src/accounts-multi-step.ts
ozsay: currently, it's very much abstract due to changes required for us in `graphql-api`.

But the idea is that the client will call it.

let's say we want to make a two-factor authentication with username-password and code received from SMS.

so the flow is like this (considering the changes here #776):

1. client will call `authenticateStep({index: 0, username: 'abc', password: '123'})` and will receive: `{ nextStep: 1, serviceId: '1234' }`
2. client will call `prepareAuthentication` from the code service.
3. after he got the SMS he can authenticate the next step:  `authenticateStep({index: 1, code: '4455', serviceId: '1234'})` and will receive the loginToken: `{serviceId: '1234', token: '...'}`
4. then he can perform the regular `loginWithService`.

davidyaha: Ah okay so this way we the client is actually oblivious to how many authentication steps there are? is there something in the response that indicates to the client that he should activate sms rather then something like an authenticator app?

ozsay: Currently, the client should be aware of the authentication steps since I can't predict here the flow of the authentication (can involve SMS sending, EMAIL...)

pradel:
@ozsay thanks for starting working on the multi-step authentication. I want this feature for quite some time inside accounts-js but never had time to start the implementation.

I took some time to try to think about the best implementation possible. Currently we have 2fa with OTP code only for passwords but the way it's done is not flexible at all and limited to the password service.

My approach was to make this multi-factor step in the core of the server package, no need to have a separate package for it. I really like the way `auth0` implemented this:
- client sends a request to the server with email + password. If the user does not have MFA enabled then return a new session. If the user has MFA enabled then return a challenge to resolve (sms, email, other...). I think we could use a graphql union to describe the return of the authenticate function (either session or challenge).
- client calls a new mutation to resolve the challenge and if the challenge is successful the server returns a new session.

You can read more about it there https://auth0.com/docs/multifactor-authentication/api/challenges.

Would love to have your thoughts about this :).

ozsay:
@pradel thanks for sharing your thoughts on this PR.

I guess that following what is done in `auth0` makes sense and seems like it may simplify clients code.

do you have any technical followup of how to do it? it will require changes in User. where do you think the current MFA state (which challenges are remaining to resolve) should be stored (locally/db)?

pradel:
@ozsay do you want to join the slack channel? Might be easier to discuss the implementation :)

ozsay:
sure why not :)

pradel:
invited

ozsay:
closing it for the better method in #796

PR #796, opened 2019-09-04
[WIP] Multi-factor authentication

ozsay:
an alternative way for #777 of doing mfa following the `auth0` way.

### DIFF
--- a/packages/database-manager/src/database-manager.ts
+++ b/packages/database-manager/src/database-manager.ts
@@ -1,15 +1,21 @@
-import { DatabaseInterface, DatabaseInterfaceSessions } from '@accounts/types';
+import {
+  DatabaseInterface,
+  DatabaseInterfaceSessions,
+  DatabaseInterfaceMfaLoginAttempts,
+} from '@accounts/types';
 
 import { Configuration } from './types/configuration';
 
 export class DatabaseManager implements DatabaseInterface {
   private userStorage: DatabaseInterface;
   private sessionStorage: DatabaseInterface | DatabaseInterfaceSessions;
+  private mfaLoginAttemptsStorage: DatabaseInterface | DatabaseInterfaceMfaLoginAttempts;
 
   constructor(configuration: Configuration) {
     this.validateConfiguration(configuration);
     this.userStorage = configuration.userStorage;
     this.sessionStorage = configuration.sessionStorage;
+    this.mfaLoginAttemptsStorage = configuration.mfaLoginAttemptsStorage;
   }
 
   private validateConfiguration(configuration: Configuration): void {
@@ -153,5 +159,17 @@ export class DatabaseManager implements DatabaseInterface {
   // Return the setResetPassword function from the userStorage
   public get setUserDeactivated(): DatabaseInterface['setUserDeactivated'] {
     return this.userStorage.setUserDeactivated.bind(this.userStorage);
+  }
+
+  public get createMfaLoginAttempt(): DatabaseInterface['createMfaLoginAttempt'] {
+    return this.mfaLoginAttemptsStorage.createMfaLoginAttempt.bind(this.mfaLoginAttemptsStorage);
+  }
+
+  public get getMfaLoginAttempt(): DatabaseInterface['getMfaLoginAttempt'] {
+    return this.mfaLoginAttemptsStorage.getMfaLoginAttempt.bind(this.mfaLoginAttemptsStorage);
+  }
+
+  public get removeMfaLoginAttempt(): DatabaseInterface['removeMfaLoginAttempt'] {
+    return this.mfaLoginAttemptsStorage.removeMfaLoginAttempt.bind(this.mfaLoginAttemptsStorage);
   }
 }
--- a/packages/database-mongo/src/mongo.ts
+++ b/packages/database-mongo/src/mongo.ts
@@ -4,10 +4,11 @@ import {
   CreateUser,
   DatabaseInterface,
   Session,
+  MfaLoginAttempt,
   User,
 } from '@accounts/types';
 import { get, merge } from 'lodash';
-import { Collection, Db, ObjectID } from 'mongodb';
+import { Collection, Db, ObjectID, MongoError } from 'mongodb';
 
 import { AccountsMongoOptions, MongoUser } from './types';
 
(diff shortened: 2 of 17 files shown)


pradel:
@ozsay since it's pretty hard to have a generic `loginWithService` and `performMfaChallenge` mutation (you can pass an email + password or a sms code or whatever for other services) what do you think about adding a JSON scalar for graphql that would allow any inputs?

packages/database-mongo/src/mongo.ts
```diff
@@ -427,4 +432,39 @@ export class Mongo implements DatabaseInterface {
...
+  public async createMfaLoginAttempt(
+    mfaToken: string,
+    loginToken: string,
+    userId: string
+  ): Promise<void> {
+    try {
+      await this.mfaLoginCollection.insertOne({ _id: mfaToken, loginToken, userId });
+    } catch (e) {
+      const me = e as MongoError;
+      if (me.code === 11000) {
```
TimMikeladze: Where is the uniqueness of the key enforced?

packages/server/src/accounts-server.ts
```diff
@@ -528,6 +577,40 @@ Please change it with a strong random token.`);
...
+
+    if (!skipValidation && loginAttempt.loginToken !== loginToken) {
+      return null;
+    }
+
+    return this.db.findUserById(loginAttempt.userId);
+  }
+
+  private async createMfaLoginProcess(user: User): Promise<MFALoginResult> {
+    const loginToken = generateRandomToken();
```
pradel: @ozsay I don't exactly understand why do we need this `loginToken`, can you explain a bit why is it really needed please? :)

ozsay: the user can only log in if he has the `loginToken` which is provided to him upon successfully completing the challenge.

Did I over-thought this? :)

pradel: Since the mfaToken is already a random token I feel like having this `loginToken` does not add more security no?

ozsay: following `auth0`, the last step of the mfa is to call `oauth/token` (in our case `loginWithService`) with a token. I can't use the `mfaToken` here because it's provided to the client before it does the challenge. that's why I have the extra token.

if we want, we can skip this step and have the `performMfaChallenge` create the session and return the access and refresh tokens.

examples/graphql-server-typescript/src/index.ts
```diff
@@ -21,6 +21,7 @@ const start = async () => {
   const accountsDb = new DatabaseManager({
     sessionStorage: userStorage,
     userStorage,
+    mfaLoginAttemptsStorage: userStorage,
```
davidyaha: Make sure `mfaLoginAttemptsStorage` is optional so that it would not break the current usages. Also no need for it when MFA is not configured

packages/graphql-api/src/modules/accounts/resolvers/loginResult.ts
```diff
@@ -1 +1,11 @@
 export const LoginResult = {};
+
+export const LoginWithServiceResult = {
+  __resolveType(obj: any) {
```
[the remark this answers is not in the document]
ozsay: not sure what you mean

packages/client-password/src/client-password.ts
```diff
@@ -25,7 +25,7 @@ export class AccountsClientPassword {
   /**
    * Log the user in with a password.
    */
-  public async login(user: any): Promise<LoginResult> {
+  public async login(user: any): Promise<LoginResult | Omit<MFALoginResult, 'mfaToken'>> {
```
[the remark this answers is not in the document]
ozsay: mainly, to simplify the usage

packages/database-manager/src/database-manager.ts
```diff
@@ -1,15 +1,21 @@
...
+  DatabaseInterfaceSessions,
+  DatabaseInterfaceMfaLoginAttempts,
+} from '@accounts/types';
 
 import { Configuration } from './types/configuration';
 
 export class DatabaseManager implements DatabaseInterface {
   private userStorage: DatabaseInterface;
   private sessionStorage: DatabaseInterface | DatabaseInterfaceSessions;
+  private mfaLoginAttemptsStorage?: DatabaseInterface | DatabaseInterfaceMfaLoginAttempts;
```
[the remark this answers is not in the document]
ozsay: @davidyaha pointed out that people might not want `mfa` at all, so I made it optional

ozsay:
@pradel 
> * Do we want to have the ability to limit the sms code to 3 attempts for example?

I think it's related to the #776 PR. The `mfa` process is kinda agnostic of which authenticators are being used.

> * should the authenticator api a part of this pull request or better to split in another one?

I don't think any changes are required to the authenticators API. I will test an alpha version this week, but I believe that it should work.

> * `mfaChallenges` as a an array of string sounds not really extendable, where will you store the number associated to the user or other?

currently, it will allow authenticating with any of the challenges in the array (only 1 challenge completion is sufficient for login). We can extend it in the future for more complex scenarios. I think the basic scenarios are covered here.

> * on the client isn't it better to return the `mfaToken` to the user instead of setting it in the local storage?

I can return it but I don't see any client usage of it, that's why I've "hidden" it inside `accounts`

thanks for the comments @pradel :)

packages/graphql-api/src/modules/accounts/resolvers/loginResult.ts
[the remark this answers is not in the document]
pradel: Instead of using any import the types generated by `graphql-codegen` :)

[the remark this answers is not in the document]
ozsay: i hope I understood you correctly :)
https://github.com/accounts-js/accounts/pull/796/commits/52ac28726f60fed6f81db73134ddc18f346284cf

acomito:
Will this allow for 2-step authentication through email, or just SMS?

davidyaha:
@ozsay Are we missing something here? Could this be merged?

pradel:
@davidyaha I proposed another design for the MFA implementation which will allow us to support more authenticators. I posted the architecture design on slack and I started an implementation on this pr https://github.com/accounts-js/accounts/pull/813

PR #798, opened 2019-09-05
Add a scalar JSON to graphql and use it for the authenticate params type

pradel:
Related to #796, @ozsay you might want to take a look.

Using a JSON scalar for the params argument will allow us to write more services without having to change the graphql schema to support that specific service


Urigo:
what do you think about using JSON from GraphQL Scalars?
https://github.com/urigo/graphql-scalars#usage

pradel:
@Urigo I saw that you were using graphql-type-json for graphql-scalars, since it's the only custom scalar we are using, I was thinking to just use it directly :)

ozsay:
@pradel 
I wonder if it can be solved using a typed way with `extend`...

I kinda solved it (prune to overrides) in https://github.com/ozsay/accounts/tree/graphql-api-2.
I basically refactored the `graphql-api` package to make it more easy to use (but it still in progress since I removed most of the configurable parts).

after the change, if I wanted to add a new service (tested for the code-service PR), I could add it to its `graphql` schema:
```
input AuthenticateParamsInput {
    serviceId: String
    code: String
}
```

The module mechanism would merge `AuthenticateParamsInput` to a unified type in the aggregated schema.

Since I'm using the same mechanism as before, I guess it works the same without my changes to `graphql-api` but I can't verify it.

pradel:
@ozsay I see, I like the idea, my only concern is that every time you add a service you will have to change the graphql schema meaning more configuration to do for the users.

pradel:
Since we can actually extend the input in graphql 14.X maybe it's better to not use JSON

pradel:
Extend is the way to go 🚀

PR #801, opened 2019-09-10
verify authentication

### LINKED ISSUE #800: Add the ability to authorize a user action
Hi,

I'm looking for an ability to authorize a user action.

some use cases:
1. A logged-in user wants to perform a sensitive action (for example, Google has the ability to prompt for fingerprint before performing an in-app purchase).
2. api request authorization, where every api request must be authorized.

besides implementing the services (for fingerprint authentication and API token validation), will be nice to have core support for this :)

Something like `authorizeWithService` (similar to `loginWithService`) can be nice.

WDYT?

ozsay:
fixes #800.

I had trouble naming the graphql & rest endpoints since the regular `loginWithService` is named `authenticate` and I didn't want to create breaking changes. I named it `authenticateWithoutSessionCreation`.

I'm open for another naming suggestions.

* After talking with @davidyaha he pointed out that `authorize` refers to roles and permissions.

Hopefully, I haven't forgotten anything


packages/graphql-api/src/modules/accounts/schema/mutation.ts
```diff
@@ -10,5 +10,6 @@ export default (config: AccountsModuleConfig) => gql`
     # Example: Login with password
     # authenticate(serviceName: "password", params: {password: "<pw>", user: {email: "<email>"}})
     authenticate(serviceName: String!, params: AuthenticateParamsInput!): LoginResult
+    authenticateWithoutSessionCreation(serviceName: String!, params: AuthenticateParamsInput!): Boolean
```
TimMikeladze: `noSession` could be add as an optional argument to the `authenticate` mutation if you want to avoid creating this extra `authenticateWithoutSessionCreation` names

ozsay: I thought about it but it will make the `loginWithService` function more complex as it would return `LoginResult | boolean`.

davidyaha:
@ozsay How about `verifyAuthentication`?

Issue #802, opened 2019-09-10
localStorage is not defined - with SSR (NextJS)


stolinski:
Trying to use with NextJS and I'm getting "localStorage is not defined". Should be an easy enough fix. I can try and submit a PR if you would like

```
ReferenceError: localStorage is not defined
    at Object.<anonymous> (/Users/scotttolinski/Sites/levelup/node_modules/apollo-accounts-password-client/node_modules/@accounts/client/lib/token-storage-local.js:49:39)
```

agustif:
@stolinski Please do!! I'm currently looking to do the same thing, would really appreciate your PR! I'm sure @pradel will be happy to review and merge it too! Could make a nice example, and be on the docs/homepage.

agustif:
this are the most relevant changes for CookieStorage to work. I think.

<img width="554" alt="Screenshot 2019-09-16 at 19 35 41" src="https://user-images.githubusercontent.com/6601142/64980001-3e764b00-d8b9-11e9-825e-03ace38df4c8.png">
<img width="755" alt="Screenshot 2019-09-16 at 19 33 51" src="https://user-images.githubusercontent.com/6601142/64980004-3e764b00-d8b9-11e9-82c6-80d3fd4b8016.png">
<img width="778" alt="Screenshot 2019-09-16 at 19 33 39" src="https://user-images.githubusercontent.com/6601142/64980005-3e764b00-d8b9-11e9-9a53-ff87b5c7b60e.png">

NickBolles:
What's the fix here? isn't it just documentation to pass in your own token storage if you're using SSR? maybe the client should check if localstorage is defined so it doesn't error out?

stolinski:
The solution would be to just check to see if localStorage exists before calling it. Unfortunately, I haven't had the time to dive in yet as I was unable to get started with the repos locally following the getting started docs. I need to find some time to dedicate to figuring that out.

agustif:
> The solution would be to just check to see if localStorage exists before calling it. Unfortunately, I haven't had the time to dive in yet as I was unable to get started with the repos locally following the getting started docs. I need to find some time to dedicate to figuring that out.

I’m on a cruiseship in a transatlantic trip, so my wifi sux big time! But please tell me how can I helo you @stolinski. I’ve had to use npm install and not yarn in some of the packages for them to work in my local machine. And also use yarn link to use some local packages like @accounts/client to tedt the nextJS example with the cookie storage ;)

edw19:
[...]

PR #804, opened 2019-09-14
NextJS Example using cookieStorage [WIP]

agustif:
**AccountsJS + NextJS (SSR)**
related to issue #802 
This is very much a WIP

Although, it seems Home/Login/Signup work as it is.

It's using apollo-link, I haven't had any problems with localStorage for now.

Still need to work out dynamic routing for reset_password. two_factor. and verify_token.

Feedback very much appreciated @pradel @TimMikeladze @davidyaha


examples/next-graphql-typescript/pages/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+      </Link>
+      <Button variant="outlined" color="primary" onClick={onLogout}>
+        <a>Logout</a>
+      </Button>
+    </Container>
+  </Layout>
+);
+
+HomePage.getInitialProps = async function(ctx) {
+  const token = auth(ctx);
```
TimMikeladze: Once you have the access token you need to called `accountsServer.resumeSession` which will verify the token, the session, and return the authenticated user. This is similar to what the context builder does in our graphql package.



You can find an example of that code here https://github.com/accounts-js/accounts/blob/6fdec832aed3e9711ea063f26bf88a45ae165e19/packages/graphql-api/src/utils/context-builder.ts#L5

pradel:
> Could you expand this example to also include an apollo client which uses the accounts apollo link to send the tokens to the app's graphql server.

@TimMikeladze that's a good idea, maybe our examples should contain some custom queries / mutations to show how you can integrate it with your code (something simple, maybe a todo list?)

TimMikeladze:
@agustif I pushed some additional code to add a graphql endpoint and create an accounts server with typeorm. It's on the https://github.com/accounts-js/accounts/tree/example-nextjs branch.

I'm not quite sure how to go about pushing changes directly to this PR... @pradel any ideas?

pradel:
@TimMikeladze I am using https://desktop.github.com/ where you have a gui to clone a pr directly on your machine and then you can push on the same pr :)

agustif:
I accepted all incoming changes @tim so yarn.lock wouldn't bitch! seems OK now!

stolinski:
This is awesome. Thank you. I can't wait to dive into this pr.

TrillCyborg:
I was having some trouble getting the TokenStorage to work in this example so I implemented my own. Hopefully this will help someone else who ran into this issue as well.

```ts
import cookies from 'next-cookies'
import Cookies from 'js-cookie'

export const tokenStorage = (ctx: any) => ({
  setItem: async (key: string, value: string) => {
    Cookies.set(key, value)
  },
  getItem: async (key: string) => {
    const allCookies = cookies(ctx)
    const item = allCookies[escape(key)]
    return item
  },
  removeItem: async (key: string) => {
    Cookies.remove(key)
  }
})
```

pradel:
@TrillCyborg I am curious, how do you do to inject the token storage to accounts-js afterwards (I guess inside getInitialProps right)?

TrillCyborg:
> @TrillCyborg I am curious, how do you do to inject the token storage to accounts-js afterwards (I guess inside getInitialProps right)?

Precisely. I pass the `ctx` in `getInitialProps` to `tokenStorage`. Since the app is SSR the Apollo/Accounts client needs to be created on each request.

agustif:
Hey could someone maybe help wrap this up?

Would love to see it merge, but I'm not sure what's missing by @TimMikeladze requsets also maybe you @pradel can help?

Thank you guys

stolinski:
I'm working on my own implementation on this and am hung up on a few things. Next recommends a HOC approach. 

https://github.com/zeit/next.js/blob/canary/examples/api-routes-apollo-server-and-client/apollo/client.js

Here, they use an initApolloClient function to initialize the client. 

I was able to get user accounts working with SSR using this approach, https://github.com/zeit/next.js/blob/canary/examples/with-apollo-auth/lib/apollo.js
However not with accounts-link. Besides the big bug in accounts-link causing an infinite loop, there is a lot of messy-ness around the load order of things. 

Has anyone unlocked this approach completely?

TrillCyborg:
> I'm working on my own implementation on this and am hung up on a few things. Next recommends a HOC approach. 
> 
> https://github.com/zeit/next.js/blob/canary/examples/api-routes-apollo-server-and-client/apollo/client.js
> 
> Here, they use an initApolloClient function to initialize the client. 
> 
> I was able to get user accounts working with SSR using this approach, https://github.com/zeit/next.js/blob/canary/examples/with-apollo-auth/lib/apollo.js
> However not with accounts-link. Besides the big bug in accounts-link causing an infinite loop, there is a lot of messy-ness around the load order of things. 
> 
> Has anyone unlocked this approach completely? 


I have accounts.js working great in a next.js project using this HOC method. What messy-ness are you running into?

stolinski:
> > I'm working on my own implementation on this and am hung up on a few things. Next recommends a HOC approach.
> > https://github.com/zeit/next.js/blob/canary/examples/api-routes-apollo-server-and-client/apollo/client.js
> > Here, they use an initApolloClient function to initialize the client.
> > I was able to get user accounts working with SSR using this approach, https://github.com/zeit/next.js/blob/canary/examples/with-apollo-auth/lib/apollo.js
> > However not with accounts-link. Besides the big bug in accounts-link causing an infinite loop, there is a lot of messy-ness around the load order of things.
> > Has anyone unlocked this approach completely?
> 
> I have accounts.js working great in a next.js project using this HOC method. What messy-ness are you running into?

I'd love to check out your implementation, or at least your withApollo. I have things a bit more worked out now, but I'm having a hard time getting the session refreshed without using accounts-link. I'll clean my up a bit and post it in a gist here.

stolinski:
@TrillCyborg After hacking away on it today, I'm running into issues regarding the order of initializing things where `GraphQLClient` requires `apolloClient` to be initialized, but `apolloClient` is only initialized within `withApollo`, however since `accountsPassword` needs to be exported, the whole thing becomes circular. I guess at the end of the day, I'm not able to get the loading order correct

TrillCyborg:
> @TrillCyborg After hacking away on it today, I'm running into issues regarding the order of initializing things where `GraphQLClient` requires `apolloClient` to be initialized, but `apolloClient` is only initialized within `withApollo`, however since `accountsPassword` needs to be exported, the whole thing becomes circular. I guess at the end of the day, I'm not able to get the loading order correct

This is what's working for me. https://gist.github.com/TrillCyborg/7ce2ba18e89176cbff02f1b86a262ca1

stolinski:
Thank so much for this. @TrillCyborg moving everything to react context was the missing piece of the loading order puzzle.

stolinski:
@TrillCyborg all is working well with this approach and I've implemented exactly, however because `opts.ctx` is undefined on creation of the accounts, I'm getting `TypeError: Cannot read property 'req' of undefined
    at nextCookies (index.js:8)` when using the tokenStorage to log in.

It seems like when the React context is created  client side , the `accountsGraphQL, accountsPassword, accountsClient` are created with ctx empty. What was your solution for this?

TrillCyborg:
> @TrillCyborg all is working well with this approach and I've implemented exactly, however because `opts.ctx` is undefined on creation of the accounts, I'm getting `TypeError: Cannot read property 'req' of undefined
>     at nextCookies (index.js:8)` when using the tokenStorage to log in.
> 
> It seems like when the React context is created  client side , the `accountsGraphQL, accountsPassword, accountsClient` are created with ctx empty. What was your solution for this?

Are you using the withApollo HOC on all your pages talking to Apollo or accounts.js?

Since the HOC uses getInitialProps you need it on components with a default export in the pages directory.

stolinski:
@TrillCyborg Yah, wrapping pages like...

```
const Home = () => {
  return (
    <MainLayout>
      <Hero />
    </MainLayout>
  )
}

export default withApollo(Home)
```
All mutations give same error. 

Before I had deleted the previously saved cookies, the code was SSRing the logged in user just fine, so the server side loading of things work fine. 

ctx is defined on server side but not client for some reason.

stolinski:
I don't know why I didn't think of this sooner, but I just adjusted the getItem method.
```
import cookies from 'next-cookies'
import Cookies from 'js-cookie'

export const tokenStorage = (ctx: any) => ({
  setItem: async (key: string, value: string) => {
    Cookies.set(key, value)
  },
  getItem: async (key: string) => {
    let allCookies
    if (typeof window !== 'undefined') {
      allCookies = Cookies.get()
    } else {
      allCookies = cookies(ctx)
    }
    const item = allCookies[escape(key)] || allCookies[key]
    return item
  },
  removeItem: async (key: string) => {
    Cookies.remove(key)
  },
  getItemClientSync: (key: string) => {
    return Cookies.get(key)
  },
})
```

edw19:
Is there a better way to use a HOC for JS accounts with the new apollo configuration? 
https://github.com/vercel/next.js/blob/canary/examples/api-routes-apollo-server-and-client/apollo/client.js
with this code

jocon128:
Hi All,

Does anyone have a recent example for using accounts-js client with nextjs?

PR #809, opened 2019-09-20
Migrate from tslint to eslint

pradel:
tslint is deprecated for eslint


pradel:
@TimMikeladze woops, updated 👍

PR #812, opened 2019-09-23
[discuss] Asymmetric authentication

ozsay:
this is a new authenticator that will utilize asymmetric cryptography for authenticating users.

The main use case is biometric authentication, following (this method](https://android-developers.googleblog.com/2015/10/new-in-android-samples-authenticating.html).

Another use case may be authenticating by PIN where you don't want to expose the PIN to the server (the PIN is being treated as a private key).

@davidyaha


packages/asymmetric/README.md
```diff
@@ -0,0 +1,30 @@
+# @accounts/password
```
ozsay: need to change that

ozsay:
@pradel I don't see the `circleci` check or the github check.

is something not working?

pradel:
Can you try to pull the account-js master branch into yours? (I think you are behind so you don't have the latest github ci config)

ozsay:
I'm only ahead of master and i can see the files: https://github.com/ozsay/accounts/tree/asymmetric-authentication/.github/workflows

pradel:
> I'm only ahead of master and i can see the files: [ozsay/accounts:.github/workflows@`asymmetric-authentication`](https://github.com/ozsay/accounts/tree/asymmetric-authentication/.github/workflows) [](/ozsay/accounts/tree/HEAD@{2019-09-23T15:51:04Z}/.github/workflows)

I wrote to the GitHub support, looks like this is happening in every pr submitted from a fork...

davidyaha:
@ozsay @pradel Can we merge this?

PR #813, opened 2019-09-23
WIP feat(mfa): new authenticator api

pradel:
I will write a description regarding the API soon, for now I just want to see if this way could work for us, related to #796.
This pr will contain the code related to the core implementation and an otp authenticator to show how it's currently working.

- [ ] Documentation
- [ ] Working rest example


packages/database-mongo/src/types/index.ts
```diff
@@ -7,6 +7,14 @@ export interface AccountsMongoOptions {
...
    */
   sessionCollectionName?: string;
+  /**
+   * The authenticators collection name, default 'authenticators'.
+   */
+  authenticatorCollectionName?: string;
+  /**
+   * The mfa challenges collection name, default 'mfaChallenges'.
+   */
+  mfaChallengeCollection?: string;
```
pradel, line 17: should be `mfaChallengeCollectionName`

Issue #814, opened 2019-09-24
How to merge with an existing GrapQL server?


bbenoist:
Hi,

I'm struggling to get an account-js microservice (GraphQL) integrated into my existing GraphQL server according to the documentation [_Usage as a GraphQL microservice_](https://accounts-js.netlify.com/docs/getting-started#usage-as-a-graphql-microservice).

First part is OK. I have an accounts-js microservice working.
On the other-side, I've tried to merge the accounts-js microservice with my existing one as follows (important parts are annotated with `********************`):

```ts
import { ApolloServer } from "apollo-server-express";
import Express from "express";
import {
  buildSchema,
  BuildSchemaOptions,
  GeneratingSchemaError
} from "type-graphql";
import { createConnection } from "typeorm";
import accountsBoost from "@accounts/boost";
import {
  makeExecutableSchema,
  mergeSchemas,
  makeRemoteExecutableSchema,
  introspectSchema,
} from "graphql-tools";
import { HttpLink } from "apollo-link-http";
import fetch from "cross-fetch";
import { setContext } from "apollo-link-context";
import { SomeResolver } from "./resolvers/SomeResolver";
import { GraphQLSchema } from "graphql";

async function initAuth() {
  const accounts = (await accountsBoost({
    tokenSecret: "my-token",
    // Setting micro to true will instruct `@accounts/boost` to only
    // verify access tokens without any additional session logic.
    micro: true,
  })).graphql();
  // Creates a link to fetch the remote schema from the accounts microservice.
  const http = new HttpLink({ uri: "http://localhost:3000", fetch });
  const link = setContext((_, previousContext) => {
    if (!previousContext.graphqlContext) return {};
    // Attach the Authorization to requests which are proxied to the remote schema.
    return {
      headers: {
        Authorization: "Bearer " + previousContext.graphqlContext.authToken,
      },
    };
  }).concat(http);
  const remoteSchema = await introspectSchema(link);
  const executableRemoteSchema = makeRemoteExecutableSchema({
    schema: remoteSchema,
    link,
  });

  // *****************************************************************
  // What am I supposed to write here?
  // *****************************************************************
  const typeDefs = `
    type PrivateType @auth {
      privateField: String
    }
    type Query {
      publicField: String
      privateField: String @auth
      privateType: PrivateType
      privateFieldWithAuthResolver: String
    }
    `;

  // *****************************************************************
  // What am I supposed to write here?
  // *****************************************************************
  const resolvers = {
[...]

PR #817, opened 2019-09-26
Update the documentation from the corresponding microservice example

bbenoist:
Changes taken from file examples/accounts-boost/src/microservice/app-server.ts

See my despair as an accounts-js newcomer in #814


pradel:
Thank you @bbenoist this will help newcomers to get started more easily 🙏

PR #825, opened 2019-10-03
fix typings for AccountsClientPassword

### LINKED ISSUE #818: Client-password typings incorrect
The parameters for [createUser](https://github.com/accounts-js/accounts/blob/master/packages/client-password/src/client-password.ts#L20) and maybe [login](https://github.com/accounts-js/accounts/blob/master/packages/client-password/src/client-password.ts#L28), should be updated to be [PasswordCreateUserType](https://github.com/accounts-js/accounts/blob/master/packages/password/src/types/password-create-user-type.ts#L4) and maybe [PasswordLoginType](https://github.com/accounts-js/accounts/blob/master/packages/password/src/types/password-login-type.ts#L4) respectively?

createUser doesn't fail because of the index type on the CreateUser type. But it's probably more correct.

NickBolles:
Closes #818

### DIFF
--- a/packages/client-password/src/client-password.ts
+++ b/packages/client-password/src/client-password.ts
@@ -1,5 +1,6 @@
 import { AccountsClient } from '@accounts/client';
-import { LoginResult, CreateUser } from '@accounts/types';
+import { LoginResult } from '@accounts/types';
+import { PasswordCreateUserType, PasswordLoginType } from '@accounts/password';
 import { AccountsClientPasswordOptions } from './types';
 
 export class AccountsClientPassword {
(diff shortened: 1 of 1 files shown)


NickBolles:
@pradel any idea why this is failing the checks? it says no linked package "@accounts/password"... shouldn't lerna be linking them all up correctly? build works locally for me

packages/client-password/src/client-password.ts
```diff
@@ -17,16 +18,24 @@ export class AccountsClientPassword {
   /**
    * Create a new user.
    */
-  public async createUser(user: CreateUser): Promise<string> {
-    const hashedPassword = this.hashPassword(user.password);
+  public async createUser(user: PasswordCreateUserType): Promise<string> {
+    let hashedPassword = user.password;
+    if (typeof user.password === 'string') {
+      // mostly just a type check
+      hashedPassword = this.hashPassword(user.password);
```
quaj: does it is necessary? server stil throw a error if user.password is number, booolean

NickBolles: The point of the PR is to fix a possibly null error I believe. It's been a while. But if that's the case hashPassword typings should allow non-string inputs then, which seems a little more messy to me.

NickBolles: Right, so the point of this fix is to fix a typescript compile issue. Honestly this fix might not be needed. I forget why I was running into it. Maybe it was just a warning

pradel:
Fixed by #893

PR #828, opened 2019-10-04
feat(accounts-password): send verification email after signup

sakulstra:
I was wondering why the verification email isn't sent automatically after signup.
At least on accounts-password I guess this should be the default behavior?

~On a side note I don't fully understand why `email` is optional here :thinking:, shouldn't the email always be present in email/password signup?~ I see, you can signup without an email, by providing a username only.

### DIFF
--- a/packages/graphql-api/src/modules/accounts-password/resolvers/mutation.ts
+++ b/packages/graphql-api/src/modules/accounts-password/resolvers/mutation.ts
@@ -16,12 +16,8 @@ export const Mutation: MutationResolvers<ModuleContext<AccountsModuleContext>> =
   },
   createUser: async (_, { user }, { injector }) => {
     const userId = await injector.get(AccountsPassword).createUser(user as PasswordCreateUserType);
-    const noErrors = injector.get(AccountsServer).options.ambiguousErrorMessages;
-    if (noErrors) {
-      user.email && (await injector.get(AccountsPassword).sendVerificationEmail(user.email));
-      return null;
-    }
-    return userId;
+    user.email && (await injector.get(AccountsPassword).sendVerificationEmail(user.email));
+    return injector.get(AccountsServer).options.ambiguousErrorMessages ? null : userId;
   },
   twoFactorSet: async (_, { code, secret }, { user, injector }) => {
     // Make sure user is logged in
--- a/packages/graphql-api/src/modules/accounts-password/resolvers/mutation.ts
+++ b/packages/graphql-api/src/modules/accounts-password/resolvers/mutation.ts
@@ -16,8 +16,12 @@ export const Mutation: MutationResolvers<ModuleContext<AccountsModuleContext>> =
   },
   createUser: async (_, { user }, { injector }) => {
     const userId = await injector.get(AccountsPassword).createUser(user as PasswordCreateUserType);
-    user.email && (await injector.get(AccountsPassword).sendVerificationEmail(user.email));
-    return injector.get(AccountsServer).options.ambiguousErrorMessages ? null : userId;
+    const noErrors = injector.get(AccountsServer).options.ambiguousErrorMessages;
+    if (noErrors) {
+      user.email && (await injector.get(AccountsPassword).sendVerificationEmail(user.email));
+      return null;
+    }
+    return userId;
   },
   twoFactorSet: async (_, { code, secret }, { user, injector }) => {
     // Make sure user is logged in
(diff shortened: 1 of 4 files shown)


packages/graphql-api/src/modules/accounts-password/resolvers/mutation.ts
```diff
@@ -16,7 +16,12 @@ export const Mutation: MutationResolvers<ModuleContext<AccountsModuleContext>> =
   },
   createUser: async (_, { user }, { injector }) => {
     const userId = await injector.get(AccountsPassword).createUser(user as PasswordCreateUserType);
-    return injector.get(AccountsServer).options.ambiguousErrorMessages ? null : userId;
+    const noErrors = injector.get(AccountsServer).options.ambiguousErrorMessages;
```
sakulstra: after looking trough the source `ambiguousErrorMessages` is not what I expected it to be :sweat_smile: 

As i get it it's just a flag to disguise error messages?

Could someone explain the current behavior to me? As I get it the implementation here is not actually related to any error, but instead just hides the registration success when `ambiguousErrorMessages` is set to true. Why?
```js
return injector.get(AccountsServer).options.ambiguousErrorMessages ? null : userId;
```

TimMikeladze: Hey @sakulstra, there's an explanation of the reasoning behind this behavior here https://github.com/accounts-js/accounts/pull/525#issuecomment-447271280 

> The createUser function should fail silently if the email we try to create already exist, so no id returned. If I am an attacker and I want to discover which user registered to your service I will call the createUser endpoint with a random email. If I get the id back means the email does not exist in the system, but if the function return null means the user exist.
If we never return the id to the client an attacker will not be able to guess if the user exist or not.

sakulstra: @TimMikeladze thanks for giving me the context :+1:
I tested it in it's current form and if i'm not misunderstanding the purpose it just doesn't work atm.

We `could` probably fix it by just wrapping
```
    const userId = await injector.get(AccountsPassword).createUser(user as PasswordCreateUserType);
```
in a `try ... catch` and only forward the error if `ambiguousErrorMessages` set to false, but i'd personally prefer just dropping it.

packages/graphql-api/src/modules/accounts-password/resolvers/mutation.ts
```diff
@@ -16,6 +16,7 @@ export const Mutation: MutationResolvers<ModuleContext<AccountsModuleContext>> =
   },
   createUser: async (_, { user }, { injector }) => {
     const userId = await injector.get(AccountsPassword).createUser(user as PasswordCreateUserType);
+    user.email && (await injector.get(AccountsPassword).sendVerificationEmail(user.email));
```
pradel: @sakulstra I can see a few problems of doing this here:
- we will need to apply the change to every storage (eg: REST).
- if you call the function directly from your server, this won't be triggered.

I guess the best option is to add this on the server under a flag.

sakulstra: Thanks for the feedback!

> add this on the server
wasn't sure if you meant accounts-server of the server part of accounts password? I guessed accounts-password as email validation might not make a lot of sense for other authentication services?

please rereview if you got some time :)

packages/password/src/accounts-password.ts
```diff
@@ -91,6 +91,10 @@ export interface AccountsPasswordOptions {
    * This function will be called when you call `createUser`.
    */
   validateUsername?(username?: string): boolean;
+  /**
```
pradel: Since this option is a boolean, can you please move it before the options?
Also, can we add a second line comment with `Default to false.`?

sakulstra: I moved it to the other booleans. Sorry for not getting back at this earlier - went on vacation and completely forgot about this as I implemented it manually in our app :/

Wondering if we should default to true though and make this a breaking change.
I guess most users using email/password auth *want* verification mails - afaik in meteor/auth this is the default behavior as well.

Anyhow, we could still do this in a later step.

sakulstra:
@pradel what's the status here, anything i can do?

pradel:
@sakulstra sorry pretty busy weeks for now, I haven't had the chance to take a look, will try to do it this week!

pradel:
Looks good thanks @sakulstra 👍

Issue #542, opened 2018-12-31
Unable to set public claims for the generated user token


haydermabood:
After checking the code I found that its not possible to set public claims for the generated tokens. Currnely `jwt.sign` received static list of data attributes: 

`
 const jwtData: JwtData = {
      token,
      isImpersonated,
      userId,
    };
`

What I want to do is to set some specific public claims according to user data, for example, 

`
const additionalClaims = {
  premiumAccount: true
};
`

Then, use those values inside resolvers logic.

It would be great if the framework can provide some kind of hook to provide additional token claim context and access those values within resolvers.

Thanks

orinokai:
I'm also very interested in this feature. I have extended the User type to add a 'role' field and it would be great to add this as a claim in the JWT token. This would allow authorization on other services simply by decoding the token and checking the user role, without having to call getUser on the authentication server. Thanks!

gbouteiller:
Any news about it?

PR #853, opened 2019-11-20
feat(server): allow tokenSecret to accept a pair of public / private key

huikaihoo:
Our project want to use RS256 in jwt signing, which use a pair of public / private key instead of a single string secret, and also we want to override `createTokens` function to add custom logic. So making the following changes in this pull request:

1. Allow `tokenSecret` to accept a pair of public / private key instead of a single string secret, and add two private function `getSecretOrPublicKey` and `getSecretOrPrivateKey` to get string secret or public / private key when sign or verify jwt.
2. ~~Make `createTokens` function become async, so when override `createTokens` function, async functions can be easily added.~~

### DIFF
--- a/examples/accounts-boost/src/monolithic/app-server.ts
+++ b/examples/accounts-boost/src/monolithic/app-server.ts
@@ -3,9 +3,11 @@ import { mergeTypeDefs, mergeResolvers } from 'graphql-toolkit';
 import { ApolloServer } from 'apollo-server';
 
 (async () => {
-  const accounts = (await accountsBoost({
-    tokenSecret: 'terrible secret',
-  })).graphql();
+  const accounts = (
+    await accountsBoost({
+      tokenSecret: 'terrible secret',
+    })
+  ).graphql();
 
   const typeDefs = `
     type PrivateType @auth {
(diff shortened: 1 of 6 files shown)


huikaihoo:
Any comment on this?

pradel:
@huikaihoo sorry for the delay I will try to take a look at it soon!

pradel:
@huikaihoo 

I just took a look at the pr and I think it would make sense to split it in 2 pr's.

> 1. Allow tokenSecret to accept a pair of public / private key instead of a single string secret, and add two private function getSecretOrPublicKey and getSecretOrPrivateKey to get string secret or public / private key when sign or verify jwt.

This sounds like a great idea indeed :+1: 

> 2. Make createTokens function become async, so when override createTokens function, async functions can be easily added.

I think this one would need more thinking, it would be great to provide an option for the user to customize what is saved inside the jwt. See https://github.com/accounts-js/accounts/issues/542.

huikaihoo:
@pradel 
So is it okay that I reverse the changes of the second point, only leave the changes of first point in this pr, and then you may merge it first?

pradel:
Sounds great!

huikaihoo:
@pradel
Reverse the changes of the second point, please check

pradel:
Thanks @huikaihoo!

PR #862, opened 2019-12-01
Fix broken docs links

gezquinndesign:
- add additional replace function
- new replace prepends the generated api docs links with /docs/api/{dir}/
- update dependencies that don't cause breaking changes


pradel:
Amazing, thanks @gezquinndesign !

PR #878, opened 2020-01-14
refactor: split database types based on services

pradel:
## ⚠️ Breaking change

Still WIP for now

- the code will be more maintainable
- easier to implement features that target only one database (can add more later)


agustif:
I've had good experiencies with sqlite, maybe could be a great way to test database code with an embeded db like sqlite instead of mongo or pg?

pradel:
@agustif the best would be to make typeorm work with sqlite I guess, what do you think?

agustif:
> @agustif the best would be to make typeorm work with sqlite I guess, what do you think?

Yeah that makes sense, I've just done a reversal of that to my codeebase from sqlite to postgres typeorm both.

Basically some types are named differently. We could probably maintain adapters for each db/adapter!

How should I start this?

PR #883, opened 2020-02-03
feat(server): add option to customize jwt payload of accessToken

### LINKED ISSUE #542: Unable to set public claims for the generated user token

huikaihoo:
Fix #542
Refer to #542 and #853, added jwtPayloadCreator as an new option in AccountsServerOptions for the user to customize what is saved inside the jwt of accessToken


packages/server/src/types/accounts-server-options.ts
```diff
@@ -35,4 +36,5 @@ export interface AccountsServerOptions {
    * Creates a new session token each time a user refreshes his access token
    */
   createNewSessionTokenOnRefresh?: boolean;
+  jwtPayloadCreator?: JwtPayloadCreator;
```
pradel: I guess this can be just a simple function directly, what do you think?
Also would be great to add a description to the option, for the generated documentation.

packages/server/src/accounts-server.ts
```diff
@@ -571,6 +574,13 @@ Please change it with a strong random token.`);
       : generateRandomToken();
   }
 
+  private async createJwtPayload(data: JwtData, user: User): Promise<JwtPayload> {
+    if (this.options.jwtPayloadCreator) {
+      return await this.options.jwtPayloadCreator.createPayload(data, user);
```
pradel: does it make sense to allow `jwtPayloadCreator.createPayload` to modify data?
I think doing something like 
```
return {
...await this.options.jwtPayloadCreator.createPayload(data, user)
 data
}
```
would be safer.

packages/server/src/types/accounts-server-options.ts
huikaihoo: @pradel  I have changed jwtPayloadCreator to simple function `createPayload`

packages/server/src/utils/tokens.ts
```diff
@@ -8,34 +8,20 @@ export const generateRandomToken = (length = 43): string => randomBytes(length).
 
 export const generateAccessToken = ({
   secret,
-  data,
+  payload = {},
   config,
 }: {
   secret: jwt.Secret;
-  data?: any;
+  payload?: any;
```
pradel: I think we can use `JwtPayload` here no?

packages/server/src/utils/tokens.ts
```diff
@@ -8,34 +8,20 @@ export const generateRandomToken = (length = 43): string => randomBytes(length).
...
 
 export const generateRefreshToken = ({
   secret,
-  data,
+  payload = {},
   config,
 }: {
   secret: jwt.Secret;
-  data?: any;
+  payload?: any;
```
pradel: I think we can use JwtPayload here no?

packages/server/src/utils/tokens.ts
huikaihoo: The default value of payload is empty object which is not JwtPayload

packages/server/src/utils/tokens.ts
huikaihoo: The default value of payload is empty object which is not JwtPayload

huikaihoo:
@pradel Resolved conflicts, please check

huikaihoo:
Any comment?

pradel:
@huikaihoo sorry I missed the notification, thanks will take a look soon!

packages/server/src/utils/tokens.ts
agustif: JwTPayload | {} beats any?

PR #892, opened 2020-02-13
chore(graphql-client): name the getUser mutation

davidyaha:
Hey!

This change is for readability on graphql loggers and also for conditional apollo links that rely on operation names.
I would love to see this merged quickly as it does not break anything.

Thanks!


pradel:
@davidyaha Published as v0.22.0

davidyaha:
Thanks @pradel!

Issue #891, opened 2020-02-11
Reduce steps for registering new user and authentication


nsine:
Currently to complete "new user" flow with graphql-api of `accounts-js` we have to do 3 steps:
*  `createUser`
* `authenticate`
* `getUser`

Which looks like something we can improve. And there are basically two improvements:
1) autologin user after registration. There is already a draft PR for this https://github.com/accounts-js/accounts/pull/829 which is waiting for adding flag for optional enabling this. I can pick it up
2) add `user` field to `LoginResult` type

And after implementing this we can create user and get his data with one call:
`mutation { createUser(...) }` ->
```
{
  tokens: {...},
  user: { id: 123, ... }
}
```

And ofc "authentication" flow reduces to one call instead of two as well.

Please let me know if this does make sense

pradel:
1. autologin after registration seems like a common request, would be great if we can pass a flag to the server instance and then it will return the session or return the id as it is now (should work for both REST and GraphQL).

I was thinking about something like this `createUser(user: CreateUserInput!): LoginResult | ID` and based on your server configuration one will be returned.

2. I am not sure we want to add the user to the `LoginResult` type since it will add a bit more complexity to the code, you can actually do it in your app yourself by extending the `LoginResult` type no?

nsine:
@pradel regarding 2nd - no complexity actually. I created a PR for this feature - https://github.com/accounts-js/accounts/pull/894

nsine:
@pradel and about autologin - actually i'm thinking of creating separate service method and mutation `createUserAndLogin` because adding flag which affects return-type can cause an issue with graphql-codegen. It will generate types assuming flag `doAutologin = false` (default value), and type of mutation will be `createUser: string`, while in the code it's `createUser: string | LoginResult` which will break typescript compilation.
What do you think?

pradel:
>  It will generate types assuming flag doAutologin = false (default value), and type of mutation will be createUser: string, while in the code it's createUser: string | LoginResult which will break typescript compilation.

I actually don't think so, types should be generated properly for this one :), this will be a breaking change for the graphql schema and needs to also work with rest, so the client types need to be updated :)

Also we should make sure to not allow this option to be true if `ambiguousErrorMessages` is set to true.

nsine:
@pradel how do you imagine gql-schema in such case?
i'd assume `createUser: ID | LoginResult` but that's not possible according to graphql spec.
[...]

PR #894, opened 2020-02-14
feat(graphql-api): return user data after "authenticate" request

nsine:
Targets https://github.com/accounts-js/accounts/issues/891

This PR:
* adds `user` field to `LoginResult` type to allowed access to logged in user data after login without doing extra request
* refactors `accounts` and `accounts-password` gql-modules. Moved common types to shared `core` module (otherwise it wasn't possible to use `User` type inside accounts-password module)


nsine:
@pradel can you please have a look at this PR
It's already open for more than 2 weeks so would like to hear some feedback if this is ok or should i update sth else here

pradel:
@nsine sorry I forgot about this one, will take a look 👍

PR #897, opened 2020-02-14
fix(apollo-link): fix infinite loop while refreshing the session

### LINKED ISSUE #839: [@accounts/apollo-link] accountsClient.refreshSession() causes infinite loop
This is tough one to debug. When using `authLink` I'm getting infinite loops and browser tab crashing after stopping and resuming my dev server.

I'm using authLink like this

```
const authLink = accountsLink(() => accountsClient)
const httpLink = new HttpLink({
  uri: '/graphql',
})

const cache = new InMemoryCache()
const client = new ApolloClient({
  link: from([authLink, httpLink]),
  cache,
  resolvers,
})
```

I'm not getting anything from the browser's profiler because it crashes out. If I comment out the authLink, the client loads, re-add it and everything works as expected until I stop and restart my dev server.

No errors or anything odd present on the server. 

Any thoughts here? Has only started happening since adding the authLink, but I honestly have no clue on the cause here. Any ideas on the best way to troubleshoot this?

pradel:
Close #839

### DIFF
--- a/packages/apollo-link-accounts/src/index.ts
+++ b/packages/apollo-link-accounts/src/index.ts
@@ -5,14 +5,15 @@ import { AccountsClient } from '@accounts/client';
 type AccountsClientFactory = () => AccountsClient | Promise<AccountsClient>;
 
 export const accountsLink = (accountsClientFactory: AccountsClientFactory): ApolloLink => {
-  return setContext(async (_, { headers: headersWithoutTokens }) => {
+  return setContext(async (req, { headers: headersWithoutTokens }) => {
     const accountsClient = await accountsClientFactory();
-    const tokens = await accountsClient.refreshSession();
-
     const headers = { ...headersWithoutTokens };
 
-    if (tokens) {
-      headers.Authorization = 'Bearer ' + tokens.accessToken;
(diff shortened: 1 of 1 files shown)


packages/apollo-link-accounts/src/index.ts
```diff
@@ -5,14 +5,19 @@ import { AccountsClient } from '@accounts/client';
...
-
     const headers = { ...headersWithoutTokens };
 
-    if (tokens) {
-      headers.Authorization = 'Bearer ' + tokens.accessToken;
+    /**
+     * We have to check this condition to avoid an infinite loop
+     * during the refresh token operation
+     */
+    if (req.operationName !== 'refreshTokens') {
```
thomasttvo: @pradel isn't 'refreshTokens' a string constant? is there anyway we can use that constant instead of using a raw value here?

thomasttvo: (I know you merge the PR, but I thought I should point out this one)

pradel: @thomasttvo that's a good idea since it's a constant, but I am not sure how to achieve that in the current setup 😁

PR #898, opened 2020-02-15
feat(password)!: add custom password hashing option

### LINKED ISSUE #811: Allow customization on hashing of password
_**TLDR:** Allow us to specify a custom `hash` function and a `verify` function in config_

Currently the `@accounts/password` package chose `bcrypt` as the hashing method. This is a good default, but imo we should be allowed to customize it, so that we can use `argon2` etc.

pradel:
Close #811 

You can now customise the hashing and verification of the password.
By default we use bcrypt to hash the password. We now provide the following 2 options:
- `hashPassword?: (password: string) => Promise<string>;` 
- `verifyPassword?: (password: string, hash: string) => Promise<boolean>;`

If you want to use `argon2` for example you can do the following:

```js
const argon2 = require('argon2');

const passwordOptions = {
  hashPassword: (password) => argon2.hash("password"),
  verifyPassword: (password, hash) => argon2.verify(hash, password),
};
```

## ⚠️ Breaking changes

- remove the `passwordHashAlgorithm` option. You can have the same behavior as before by providing the new `hashPassword` option.

### DIFF
--- a/packages/password/__tests__/utils/encryption.ts
+++ b/packages/password/__tests__/utils/encryption.ts
@@ -1,17 +1,10 @@
-import { bcryptPassword, hashPassword, verifyPassword } from '../../src/utils/encryption';
+import { bcryptPassword, verifyPassword } from '../../src/utils/encryption';
 
 describe('encryption', () => {
   describe('bcryptPassword', () => {
     it('should return the hashed password', () => {
       const password = 'pass';
       expect(bcryptPassword(password)).not.toBe(password);
-    });
-  });
-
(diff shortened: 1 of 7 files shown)


leesiongchan:
Should we not hardcode `bcrypt` `db.setPassword` as well? For example https://github.com/accounts-js/accounts/blob/master/packages/database-typeorm/src/typeorm.ts#L249

pradel:
@leesiongchan that's a good idea but the problem is that would require a migration, in order to avoid a breaking change I guess it's fine to leave it as it is now :)

PR #829, opened 2019-10-04
feat(autoLogin): automatically login after registration

sakulstra:
Currently i'm getting a `"message": "Invalid credentials",` error... didn't yet figure out why but will have a closer look soon.

- targets #377 for accounts/password users

PR #901, opened 2020-02-17
feat!: add option to automatically login user after registration

nsine:
Targets #891 
And continues this PR https://github.com/accounts-js/accounts/pull/829

## Breaking change
* updates return type of `createUser` mutation to new `CreateUserResult` type

This PR:
* adds autologin feature (can be enabled by setting `enableAutologin` flag to `true` in `AccountServer` config)
* updates graphql-api/client, rest-express/client packages for new `createUser` response type
* adds new `CreateUserResult` type which includes `userId` and `loginResult` fields which is filled by login data in case autologin is enabled and `null` otherwise
* cherry-picks my commit from other branch https://github.com/accounts-js/accounts/pull/894/commits/0e3ec90dad10ccf43f5fd1a0da72d3a2a438ec2c to refactor shared types in graphql-api


nsine:
@pradel can you please have a look and probably help me with writing tests for this
I tried to quickly understand how tests are organized in the app, but didn't make it 😕

pradel:
@nsine thanks! Will try to take a look before the end of the week

packages/graphql-api/src/modules/accounts-password/resolvers/mutation.ts
```diff
@@ -24,11 +24,36 @@ export const Mutation: MutationResolvers<ModuleContext<AccountsModuleContext>> =
...
+      return {
+        userId: accountsServer.options.ambiguousErrorMessages ? null : userId,
+      };
+    }
+
+    // When initializing AccountsServer we check that enableAutologin and ambiguousErrorMessages options
+    // are not enabled at the same time
+
+    const { password, ...rest } = user;
+    const loginResult = await accountsServer.loginWithService(
```
pradel: here it's probably better to call `loginWithUser` directly :)

packages/rest-express/src/endpoints/password/register.ts
```diff
@@ -1,15 +1,45 @@
...
+    }
+
+    // When initializing AccountsServer we check that enableAutologin and ambiguousErrorMessages options
+    // are not enabled at the same time
+
+    const userAgent = getUserAgent(req);
+    const ip = requestIp.getClientIp(req);
+
+    const { password, ...rest } = req.body.user;
+    const loginResult = await accountsServer.loginWithService(
```
pradel: here it's probably better to call `loginWithUser` directly :)

pradel:
For the test, first of all not of them have tests. Tests are under an `__tests__` folder, if you run the `yarn coverage` command inside one of the packages, you should see the coverage numbers displayed so just make sure that you add tests for your conditions basically :)

nsine:
@pradel i've added tests for autologin feature and applied your suggestions. Please have a look again 🙂

pradel:
@nsine thank you for the pr, that's a great addition!

sakulstra:
@pradel could we get a release with this :pray:?

agustif:
Awesome just enabled this! Thanks for the hard work !

Issue #919, opened 2020-03-29
Custom payload for OAuth user registration


nsine:
# Feature request
Custom payload when registration through Oauth service
## Is your feature request related to a problem? Please describe.
Currently when using oauth service - after authentication if user doesn't exist - he is created only with email https://github.com/accounts-js/accounts/blob/master/packages/oauth/src/accounts-oauth.ts#L39
However in some cases we need to pass additional data. E.g. many oauth providers return additional data from "access_token" endpoint which you would probably like to save (like first or last name). Or probably your user model will require some additional data to be valid (e.g. `role`)
And i didn't find not-hacky way to implement this
## Describe the solution you'd like
We can pass additional function like `getRegistrationPayload` to convert response from oauth api to registration payload or set additional fields in this method.
I expect this to work like this:
```
class AccountsOAuthTwitter {
  public authenticate(params: any) {
  }
  public getRegistrationPayload(oauthUser: OauthUser) {
    return {
      username: oauthUser.username,
      role: 'ADMIN',
      ....
    }
  }
}
```

Will create a PR soon as i already did sth like this in our project

PR #920, opened 2020-03-29
feat(oauth): allow to pass additional function to make custom payload for user creation

nsine:
closes: https://github.com/accounts-js/accounts/issues/919


packages/oauth/src/types/oauth-options.ts
```diff
@@ -2,6 +2,13 @@ import { OAuthUser } from './oauth-user';
...
 export interface OAuthOptions {
   [provider: string]: {
+    /**
+     * Should return oauth user payload from passed authentication code or token
+     */
     authenticate: (params: any) => Promise<OAuthUser>;
+    /**
+     * Allows to provide function to create custom payload for user creation method based on data from OAuth API
+     */
+    getRegistrationPayload?: (oauthUser: OAuthUser) => Promise<any>;
```
pradel, line 12: can we pass the provider name as a first param? that way you would be able to do some conditional logic based on twitter of facebook for example

pradel:
@nsine thanks for the pr!
I like the idea of allowing this function, I am just wondering if it should be in the OAuth package or rather in the subpackage (eg: twitter, facebook..).

nsine:
@pradel 
From my perspective I'm sure it should be configured per-specific provider.
This is how i'm using OAuth package:
```typescript
const accountsOauth = new AccountsOauth({
  linkedIn: {
    authenticate: async (params) => {
      // getting user from linkedin
    },
    getRegistrationPayload: (oauthUser) => ({
      otherFieldsIHaveToSet: {},
      profile: {
        firstName: oauthUser.data.localizedFirstName,
        lastName: oauthUser.data.localizedLastName,
      },
    }),
  },
});
```

which is basically equal to 
```
class AccountsOauthLinkedIn {
  authenticate() {
    //...
  }

  getRegistrationPayload(oauthUser) {
    //...
  }
}

const accountsOauth = new AccountsOauth({
  linkedIn: new AccountsOauthLinkedIn()
});
```

And I really like the way how accounts-js oauth works here so I can just pass custom class with `authenticate` function for specific oauth-provider and this will work. And together with that `getRegistrationPayload` is also provider-specific function which can be written in that custom class which is pretty I think 🙂

packages/oauth/src/types/oauth-options.ts
nsine: but we already write it inside `provider` object https://github.com/accounts-js/accounts/pull/920/files/900157dedadcb7e479e8379dad1799e9fc8364ac#diff-dc3049153df06cd271b8ee347f5852e8R4

wrote more details about this in this comment: https://github.com/accounts-js/accounts/pull/920#issuecomment-610845227

pradel: true sorry I totally missed the line 4 so I thought it was directly passed as an option 👍

pradel:
@nsine the idea is good, but currently this won't work with what we have.
We have [oauth-twitter](https://github.com/accounts-js/accounts/tree/master/packages/oauth-twitter) and [oauth-instagram](https://github.com/accounts-js/accounts/tree/master/packages/oauth-instagram) and currently, you need to use them like this:

```ts
const accountsOauth = new AccountsOauth({
  twitter: new AccountsOauthTwitter({})
});
```
So if we want your option to work it should probably be an option for the twitter package, eg:
```ts
const accountsOauth = new AccountsOauth({
  twitter: new AccountsOauthTwitter({
    getRegistrationPayload: (oauthUser) => ({
    })
  })
});
```

nsine:
@pradel I think we didn't understand each other here 😅
What you said is exactly what I wrote in my previous comment 🙂 I just used LinkedIn provider as an example which is same as we currently have in this repo for Twitter.

I believe this misunderstanding is because I added this field to OAuthOptions interface which is in base `oauth` package but as you can see - https://github.com/accounts-js/accounts/pull/920/files#diff-dc3049153df06cd271b8ee347f5852e8R3 - these methods(`authenticate` and `getRegistrationPayload` I added) are already for specific provider `[provider: string]` (twitter, instagram, google etc.). And what you currently have for twitter and instagram is exactly implementation of this { authenticate } interface. For better "OOP" these types should be declared as
```typescript
interface OauthProvider {
  authenticate: (params: any) => Promise<OAuthUser>;
  getRegistrationPayload?: (oauthUser: OAuthUser) => Promise<any>;
}

export interface OAuthOptions {
  [provider: string]: OauthProvider;
}
```
(I can improve this, but out from this PR)

In this PR I added a possibility of writing this custom `getRegistrationPayload` method. Implementing this method for default twitter and instagram providers we have here doesn't make too much sense I think because we can't know what "User" models do account-js users have. But after this one is merged this becomes extendable. E.g. if I want to save first and last name to database after user is signed up with instagram, I will be able to write sth like this:
```typescript
class CustomInsagramOauth extends AccountsOAuthInstagram {
  getRegistrationPayload(oauthUser) {
    return { email: oauthUser.email, firstName: oauthUser.profile.firstName, lastName: oauthUser.profile.lastName };
  }
}

const accountsOauth = new AccountsOauth({
  instagram: new CustomInsagramOauth() 
});
```

hope i made this more clear 🙂

nsine:
ah, and if you also assumed using default twitter and instagram providers like this
```typescript
const accountsOauth = new AccountsOauth({
  twitter: new AccountsOauthTwitter({
    getRegistrationPayload: (oauthUser) => ({
    })
  })
});
```

I can also update them to be configurable with this function instead of class extending i wrote above. Because everything i wrote before was in assumption of writing custom oauth provider (which i did in our project), didn't think too much about existing twitter and instagram 🤷‍♂️

pradel:
> I can also update them to be configurable with this function instead of class extending i wrote above. Because everything i wrote before was in assumption of writing custom oauth provider (which i did in our project), didn't think too much about existing twitter and instagram 🤷‍♂️

Yup that was my point to have something that we can also use for the existing packages, so more as an option rather than extending the class :D

nsine:
@pradel i added this func to options for instagram and twitter providers. Please have a look again 🙏

nsine:
Also was thinking if this is better to make `OAuthProvider` to be abstract class instead of interface https://github.com/accounts-js/accounts/pull/920/files#diff-6855a55583d9fe2f7d3ae958b028df46R6 with default `getRegistrationPayload` implementation returning `{ email: oauthUser.email }` which i currently have written here https://github.com/accounts-js/accounts/pull/920/files#diff-6855a55583d9fe2f7d3ae958b028df46R6

But that would possibly break current code with custom providers because they don't extend `OAuthProvider` class which will be required. So left it to be an interface

nsine:
@pradel kind reminder to review 🙂would be cool to see it in next version if everything's ok

PR #931, opened 2020-04-07
feat(password): remove reset password tokens on change password

pfabrikant:
Adding the method `removeAllPasswordResetTokens` to the `password`  package. This method is called within the `changePassword` method. It removes all password reset tokens from the user storage.

### DIFF
--- a/packages/database-mongo/__tests__/index.ts
+++ b/packages/database-mongo/__tests__/index.ts
@@ -731,7 +731,8 @@ describe('Mongo', () => {
       const mongoOptions = new Mongo(databaseTests.db, {
         convertUserIdToMongoObjectId: false,
       });
-      await mongoOptions.setService('toto', 'twitter', { id: '1' });
+      const userId = await mongoOptions.createUser(user);
+      await mongoOptions.removeAllPasswordResetTokens(userId);
     });
 
     it('should remove the password reset tokens', async () => {
--- a/packages/database-mongo/__tests__/index.ts
+++ b/packages/database-mongo/__tests__/index.ts
@@ -3,7 +3,6 @@ import { ObjectID, ObjectId } from 'mongodb';
 
 import { Mongo } from '../src';
 import { DatabaseTests } from './database-tests';
-import { isMainThread } from 'worker_threads';
 
 const databaseTests = new DatabaseTests();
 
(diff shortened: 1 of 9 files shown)


packages/password/src/accounts-password.ts
```diff
@@ -88,6 +88,12 @@ export interface AccountsPasswordOptions {
    * Will automatically send a verification email after signup.
    * Default to false.
    */
+
+  removeAllPasswordResetTokensAfterPasswordChanged?: boolean;
+  /**
```
pradel, line 93: looks like this comment is not for the good option ;)

packages/password/__tests__/accounts-password.ts
```diff
@@ -397,10 +431,12 @@ describe('AccountsPassword', () => {
       const userId = 'id';
       const setPassword = jest.fn(() => Promise.resolve('user'));
       const findUserById = jest.fn(() => Promise.resolve(validUser));
-      password.setStore({ setPassword, findUserById } as any);
+      const removeAllPasswordResetTokens = jest.fn(() => Promise.resolve());
```
pradel, line 434: in this test, where do you check that it was actually called?

packages/database-mongo/src/mongo.ts
```diff
@@ -364,6 +364,18 @@ export class Mongo implements DatabaseInterface {
     );
   }
 
+  public async removeAllPasswordResetTokens(userId: string): Promise<void> {
+    const id = this.options.convertUserIdToMongoObjectId ? toMongoID(userId) : userId;
+    await this.collection.update(
+      { _id: id },
+      {
+        $set: {
```
pradel, line 372: I am not sure it really makes sense to set an empty array here, we could just unset the value to leave a clean object in the db, what do you think?

packages/database-mongo/__tests__/index.ts
```diff
@@ -725,6 +725,30 @@ describe('Mongo', () => {
     });
   });
 
+  describe('removeAllPasswordResetTokens', () => {
+    // eslint-disable-next-line jest/expect-expect
+    it('should not convert id', async () => {
+      const mongoOptions = new Mongo(databaseTests.db, {
+        convertUserIdToMongoObjectId: false,
+      });
+      await mongoOptions.setService('toto', 'twitter', { id: '1' });
```
pradel, line 734: you need to call `removeAllPasswordResetTokens` here 😁

packages/database-mongo/__tests__/index.ts
```diff
@@ -725,6 +725,30 @@ describe('Mongo', () => {
...
+    });
+
+    it('should remove the password reset tokens', async () => {
```
pradel, line 737: perfect test 👌

packages/password/src/accounts-password.ts
pfabrikant: oops :)

packages/password/__tests__/accounts-password.ts
pfabrikant: My mistake. I checked the wrong method in the assertion.

packages/database-mongo/__tests__/index.ts
pfabrikant: 🤦‍♂

packages/password/src/accounts-password.ts
```diff
@@ -84,6 +84,11 @@ export interface AccountsPasswordOptions {
    * Default to false.
    */
   invalidateAllSessionsAfterPasswordChanged?: boolean;
+  /**
+   * Will remove all password reset tokens from the db after a password has been changed.
+   * Default to false.
```
pradel, line 89: default to true actually :)

pfabrikant: 🤦‍♂

pradel:
Thanks @pfabrikant! Will create now

PR #932, opened 2020-04-07
feat: add new `useInternalUserObjectSanitizer` option

dotansimha:
- This might take up to 300ms, and if someone is using Accounts with express, it will run as part of the middleware.
Adding 300ms to something we don't really need the response of, is not needed.

- This PR also includes a fix for this issue: https://github.com/accounts-js/accounts/issues/933 

- Also, I pushed an optional fix for `express-session`. The `resolve` method there is not really mandatory, since we are running `renew`, which executes `refreshTokens` and eventually pulls `User` from the database, so we have the `user` already. So `resolve` could be optional!

### DIFF
--- a/packages/express-session/__tests__/accounts-session.ts
+++ b/packages/express-session/__tests__/accounts-session.ts
@@ -8,6 +8,7 @@ const mockAccountsServer = () => ({
       accessToken: 'newAccess',
       refreshToken: 'newRefresh',
     },
+    user: 'user',
   }),
   logout: jest.fn(),
 });
@@ -18,7 +19,7 @@ describe('AccountsSession', () => {
 
     expect(accountsSession.options.name).toEqual(defaultName);
     expect(accountsSession.options.user.name).toEqual('user');
-    expect(typeof accountsSession.options.user.resolve).toBe('function');
+    expect(accountsSession.options.user.resolve).toBe(null);
   });
 
   test('should should set', () => {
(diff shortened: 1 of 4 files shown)


dotansimha:
@pradel thank you for the feedback! 

I did try `express-session`, and everything seems to work, it reduces the amount of database operations, because we not longer need to refetch the session object and the user object :)  @kamilkisiela what do you think? 

Regarding `updateSession`, I can revert it if you think it might cause issues, but consider that we are updating an object that we just pulled, and most databases update operation doesn't fail when object doesn't exists. So the chance for an update operation to fail is very low. 
Also, if there is an issue with the database, it will fail anyway on an previous or next operation. 

Regarding the sanitizer, what do you think about using keep the sanitizer function as-is, but add a flag for `useInternalUserSanitizer`? it will workaround the breaking change and allow to override when it's needed.

pradel:
> I can also do Promise.all to make sure it runs in parallel with other operations, but still wait for the response of it.

Sound like a good solution :) 

> Regarding the sanitizer, what do you think about using keep the sanitizer function as-is, but add a flag for useInternalUserSanitizer? it will workaround the breaking change and allow to override when it's needed.

👍

kamilkisiela:
@pradel @dotansimha  LGTM

dotansimha:
Ok so turns out we can't have `Promise.all` because it's the only operation, other operations occur in a different flow. 
I will bring back the `await`. 

I changed the sanitizier implementation as well. 

@pradel can we get this merged and release? :)

pradel:
@dotansimha can you add the option here `packages/server/src/types/accounts-server-options.ts` with a description? Otherwise typescript will not allow you to pass the option :)

dotansimha:
done! @pradel

pradel:
Thanks @dotansimha will publish a new release later today!

pradel:
@dotansimha released https://github.com/accounts-js/accounts/releases/tag/v0.25.0

PR #934, opened 2020-04-10
fix(graphql-client): fix add email mutation

gezquinndesign:
This PR:

- fixes the addEmail feature which was incorrectly linked to changePassword in the graphql-client package
- adds  the missing addEmail mutation.
- includes a `verified` parameter on the mutation to allow users to specify if the new email they're adding should be marked as verified.
- Updates the following packages to take account of the new mutation: `client`, `client-password`, `graphql-api` and `graphql-client`
- adds some missing types that were flagged whilst running the tests

### DIFF
--- a/packages/graphql-client/src/graphql-client.ts
+++ b/packages/graphql-client/src/graphql-client.ts
@@ -8,6 +8,7 @@ import {
 } from '@accounts/types';
 import { print } from 'graphql';
 import gql from 'graphql-tag';
+import { addEmailMutation } from './graphql/add-email.mutation';
 import { authenticateWithServiceMutation } from './graphql/authenticate-with-service.mutation';
 import { changePasswordMutation } from './graphql/change-password.mutation';
 import { createUserMutation } from './graphql/create-user.mutation';
@@ -141,7 +142,7 @@ export default class GraphQLClient implements TransportInterface {
    * @inheritDoc
    */
   public async addEmail(newEmail: string): Promise<void> {
-    return this.mutate(changePasswordMutation, 'addEmail', { newEmail });
+    return this.mutate(addEmailMutation, 'addEmail', { newEmail });
   }
 
   /**
--- a/packages/graphql-client/src/graphql/add-email.mutation.ts
+++ b/packages/graphql-client/src/graphql/add-email.mutation.ts
@@ -1,0 +1,7 @@
+import gql from 'graphql-tag';
+
+export const addEmailMutation = gql`
+  mutation addEmail($newEmail: String!) {
+    addEmail(newEmail: $newEmail)
+  }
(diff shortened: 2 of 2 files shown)


nsine:
@gezquinndesign i think that would be a nice update to allow this.
But wouldn't this be a potential security hole? E.g. if in our app we are not supposed to allow users to skip email verification for new added emails - a potential hacker-user still can add it with direct api call `mutation { addEmail(newEmail: 'imhacker@g.co',  verified: true) }`

Not sure what would be proper way to handle this. May be you can add some flag like `allowEmailAutoverification`?
cc @pradel

gezquinndesign:
Not sure if I've got it exactly right but I think addEmail is only available if the user is logged in. 

Happy to remove the `verified` bit if it will create a security issue. 

I just added it because it was available in the [mongo package](https://github.com/accounts-js/accounts/blob/0d481c1243a59193bae525c1b6e7b46c34281e67/packages/database-mongo/src/mongo.ts#L166) but there was nothing client-side that allowed you to set it.

pradel:
As @nsine said I don't think it's a good idea to allow the verified field to be set by the user. An email should be verified only by clicking on the verify email link since we should be able to trust it :)

The server exposes the verified option in case you want to do some custom behavior (eg: customer support from your company will use a special admin panel to do this).

website/package.json
```diff
@@ -30,8 +30,9 @@
     ]
   },
   "devDependencies": {
-    "ts-node": "8.5.4",
```
pradel, line 33: I think these changes are not related to the pr 😁

website/yarn.lock
```diff
@@ -1401,6 +1401,11 @@
   resolved "https://registry.yarnpkg.com/@xtuc/long/-/long-4.2.2.tgz#d291c6a4e97989b5c61d9acf396ae4fe133a718d"
   integrity sha512-NuHqBY1PB/D8xU6s/thBgOAiAP7HOYDQ32+BFZILJ8ivkUkAHQnWfn6WhL79Owj1qmUnoN/YPhktdIoucipkAQ==
 
+"@yarnpkg/lockfile@^1.1.0":
```
pradel, line 1404: I think these changes are not related to the pr 😁

gezquinndesign:
OK. I'll remove `verified` and the other non-related changes and resubmit.

gezquinndesign:
OK, so I've cleaned up my PR so that it is one commit with the necessary changes required to fix the bug.  I've taken out any unrelated changes.  I rebased to the last release commit but the test suite fails on the e2e tests even without my changes - not sure if that's because there's some work in progress.

pradel:
Thanks @gezquinndesign !

Issue #938, opened 2020-04-15
accounts/graphql-client bundle size


sakulstra:
# Bug report
## Describe the bug
When adding @accounts/graphql-client we realized that our bundle size increased by nearly 150kb - it turns out to be related to how `graphql` is bundled within the package - not quite sure yet how to fix it though.

## To Reproduce
Add the dependency to your project.

## Expected behavior
not having the complete graphql package in the build/ having a smaller footprint when using this module
- Version of accounts-js: [e.g. 0.0.24]

### Additional info
So it seems like this is not only a problem with specifically the "graphql" package and most likely an issue with all the client side packages in here. The "problem" from my understanding seems to be that using tsc to compile stuff: https://github.com/accounts-js/accounts/blob/master/packages/graphql-client/package.json#L11 doesn't do fancy things like tree shaking - this leads to complete node/modules ending up in the bundle when e.g. importing `{sth} from 'graphql'`.

nsine:
tldr;
bundle size is super huge because of https://github.com/accounts-js/accounts/blob/master/packages/graphql-client/src/graphql-client.ts#L9
will create a pr to fix this

pradel:
Fixed in v0.25.1

PR #939, opened 2020-04-16
fix(graphql-client): reduce bundle size by importing only graphql subpackages

nsine:
closes: https://github.com/accounts-js/accounts/issues/938

### DIFF
--- a/packages/graphql-client/package.json
+++ b/packages/graphql-client/package.json
@@ -34,5 +34,8 @@
     "@accounts/types": "^0.25.0",
     "graphql-tag": "^2.10.0",
     "tslib": "1.10.0"
+  },
+  "peerDependencies": {
+    "graphql": "^0.11.0 || ^0.12.0 || ^0.13.0 || ^14.0.0"
   }
 }
--- a/packages/graphql-client/src/GraphQLErrorList.ts
+++ b/packages/graphql-client/src/GraphQLErrorList.ts
@@ -1,4 +1,4 @@
-import { GraphQLError } from 'graphql';
+import { GraphQLError } from 'graphql/error/GraphQLError';
 
 export class GraphQLErrorList extends Error {
   public errors: readonly GraphQLError[];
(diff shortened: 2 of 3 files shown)


packages/graphql-client/package.json
```diff
@@ -34,5 +34,8 @@
     "@accounts/types": "^0.25.0",
     "graphql-tag": "^2.10.0",
     "tslib": "1.10.0"
+  },
+  "peerDependencies": {
+    "graphql": "^0.11.0 || ^0.12.0 || ^0.13.0 || ^14.0.0"
```
nsine, line 39: also graphql-client uses graphql package (obviously 😅) but doesn't declare it as a dependency.
Added to peer dependencies (versions are compatible with apollo-client 2&3)

nsine:
Screenshots of webpack-bundle-analyzer to see the difference
Before:
image
After:
image

packages/graphql-client/package.json
pradel: @nsine good catch 😁 does it make sense to also add it as devDependency?
Also is it compatible with graphql 15?

nsine: it shouldn't be in `devDependencies` as it's required in actual client code to work, not only while development. When choosing `dependencies` or `peerDependencies`: i found that many libraries require it as peer dep. to avoid multiple packages with different versions + there is an explanation from the creator why it's better to use peerDependencies:
https://leebyron.com/graphql-js-preparing-for-v14/#duplicate-installs-and-peer-dependencies

Regarding v15 - i'm not sure to be honest 😅We don't use it yet on our project because there are some open issues with v15 + tooling packages (apollo-link, apollo-toolkit afaik). Plus even apollo client doesn't allow to use v15 yet https://github.com/apollographql/apollo-client/blob/master/package.json#L58 So i would go with <=14 for now

pradel: Ok perfect then 👍

pradel:
Thanks @nsine !

PR #943, opened 2020-04-28
feat(graphql-client): load user data after login/registration when using @accounts/graphql-client

nsine:
completes what i did in https://github.com/accounts-js/accounts/pull/894

I forgot there about those who are using accounts accounts client with graphql transport 🙊(now we do). So graphql-client's `createUser` and `authenticate` mutations responses will also contain user data => it will be available when using accounts `client` and/or `client-password`

### DIFF
--- a/packages/graphql-client/src/graphql-client.ts
+++ b/packages/graphql-client/src/graphql-client.ts
@@ -63,7 +63,7 @@ export default class GraphQLClient implements TransportInterface {
    * @memberof GraphQLClient
    */
   public async createUser(user: CreateUser): Promise<CreateUserResult> {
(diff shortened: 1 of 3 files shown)


pradel:
Thanks @nsine!

agustif:
@nsine sorry for hijacking the PR, but where would be OK to contact you about oauth implementation of AccountsJS. I need some help and maybe you can give me some tips/ideas on how to proceed? I'm trying to build out a fulll example front/bacck of oauth-facebook.

Thanks, and sorry for this being off-topic. I sent you a twit maybe DM me there if you will, or let me know if you're on AccountsJS-Slack! Didn't find you by @nsine or Vlad

PR #972, opened 2020-05-13
Alpha for create-accounts-app. Forked from create-next-app

agustif:
I was testing this out as an idea, it seems it could be useful enough in case someone would liike to download one of the example projects via a cli instead of github-ui

Also can work with separate projects as github repo's too (as input)

What do you guys think?

You can run it by 

yarn or 'npm install'

yarn build

node dist/index.js

You can add it to your path too if you like and use ite as create-accounts-app command, if you like it it could be published as npm package and installed as such.

Some little demo:

https://i.imgur.com/l7lWCmm.mp4


agustif:
Hey just double-checking. This got close due to me usong the anti-pattern of opening a PR from master instead of it’s own separate branch, right?

I always forget. Will reopen a new one properly set 😅

pradel:
@agustif not sure why pull bot closed it 🤔 we don't have it enabled on this repo, is it possible that you added it on yours?

agustif:
Yep, indeed it was me I had it enabledin settings and I didn't even remember!

Thanks for the info, will reopen later!

PR #984, opened 2020-05-25
feat(server): add option to customise JWT payload of accessToken

### LINKED ISSUE #542: Unable to set public claims for the generated user token

pradel:
Fix #542
Close #883

Created from #883

@huikaihoo I created a new pull request because I was not able to push on your branch


pradel:
Thanks @huikaihoo, sorry that this one took so long to get merged, I will release a new version later today!

PR #990, opened 2020-05-28
feature: add exclusion list to invalidateAllSessions method

arifaydogmus:
PR adds new optional parameter to invalidateAllSessions method to allow not destroy all session on specific cases. (Such as invalidate all sessions except active one)

### DIFF
--- a/packages/database-mongo/src/mongo.ts
+++ b/packages/database-mongo/src/mongo.ts
@@ -368,25 +368,20 @@ export class Mongo implements DatabaseInterface {
     );
   }
 
-  public async invalidateAllSessions(userId: string, excludeList?: string[]): Promise<void> {
-    let selector;
-    selector = { userId };
+  public async invalidateAllSessions(userId: string, excludedSessionIds?: string[]): Promise<void> {
+    const selector: { userId: string; _id?: object } = { userId };
 
-    if (excludeList && excludeList.length > 0) {
-      let excludedObjectIds;
-      excludedObjectIds = excludeList;
+    if (excludedSessionIds && excludedSessionIds.length > 0) {
+      let excludedObjectIds: string[] | ObjectID[] = excludedSessionIds;
 
       if (this.options.convertSessionIdToMongoObjectId) {
-        excludedObjectIds = excludeList.map((sessionId) => {
+        excludedObjectIds = excludedSessionIds.map((sessionId) => {
           return toMongoID(sessionId);
         });
       }
 
-      selector = {
-        userId,
-        _id: {
-          $nin: excludedObjectIds,
-        },
+      selector._id = {
+        $nin: excludedObjectIds,
       };
     }
 
--- a/packages/database-mongo/src/mongo.ts
+++ b/packages/database-mongo/src/mongo.ts
@@ -368,16 +368,34 @@ export class Mongo implements DatabaseInterface {
     );
   }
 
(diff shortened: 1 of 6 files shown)


packages/types/src/types/session/database-interface.ts
```diff
@@ -21,5 +21,5 @@ export interface DatabaseInterfaceSessions {
 
   invalidateSession(sessionId: string): Promise<void>;
 
-  invalidateAllSessions(userId: string): Promise<void>;
+  invalidateAllSessions(userId: string, excludeList?: string[]): Promise<void>;
```
pradel, line 24: What do you think about having a self-explanatory name instead of something generic (eg: excludeSessionIds or smth like this)?

packages/database-mongo/src/mongo.ts
```diff
@@ -368,16 +368,34 @@ export class Mongo implements DatabaseInterface {
...
-  public async invalidateAllSessions(userId: string): Promise<void> {
-    await this.sessionCollection.updateMany(
-      { userId },
-      {
-        $set: {
-          valid: false,
-          [this.options.timestamps.updatedAt]: this.options.dateProvider(),
-        },
+  public async invalidateAllSessions(userId: string, excludeList?: string[]): Promise<void> {
+    let selector;
```
pradel, line 372: Would be great to type the selector so it's not defaulting to any :)

packages/database-typeorm/src/typeorm.ts
```diff
@@ -422,14 +422,19 @@ export class AccountsTypeorm implements DatabaseInterface {
     }
   }
 
-  public async invalidateAllSessions(userId: string): Promise<void> {
-    await this.sessionRepository.update(
-      {
+  public async invalidateAllSessions(userId: string, excludeList?: string[]): Promise<void> {
+    let selector;
```
pradel, line 426: Would be great to type the selector so it's not defaulting to any :)

packages/database-mongo/src/mongo.ts
```diff
@@ -368,16 +368,34 @@ export class Mongo implements DatabaseInterface {
...
+      excludedObjectIds = excludeList;
+
+      if (this.options.convertSessionIdToMongoObjectId) {
+        excludedObjectIds = excludeList.map((sessionId) => {
+          return toMongoID(sessionId);
+        });
       }
-    );
+
+      selector = {
```
pradel, line 385: to avoid conflict with future stuff we might add later the best would be to do `selector._id = ...`. That way we do not overwrite the previous object and can keep the old conditions

packages/database-mongo/src/mongo.ts
```diff
@@ -368,16 +368,34 @@ export class Mongo implements DatabaseInterface {
...
-        $set: {
-          valid: false,
-          [this.options.timestamps.updatedAt]: this.options.dateProvider(),
-        },
+  public async invalidateAllSessions(userId: string, excludeList?: string[]): Promise<void> {
+    let selector;
+    selector = { userId };
+
+    if (excludeList && excludeList.length > 0) {
+      let excludedObjectIds;
```
pradel, line 376: Can we type this one?

packages/types/src/types/session/database-interface.ts
arifaydogmus: Yup. make sense. 👍

Issue #991, opened 2020-05-28
`DataTypeNotSupportedError` in `graphql-server-typeorm-postgres`


idkjs:
# Bug report

Running the [postgres example](https://github.com/accounts-js/accounts/blob/master/examples/graphql-server-typeorm-postgres/README.md) throws error described below.

## Describe the bug

```sh
❯ yarn start-only
yarn run v1.22.4
$ NODE_ENV=development nodemon -w src -x ts-node src/index.ts
[nodemon] 2.0.3
[nodemon] to restart at any time, enter `rs`
[nodemon] watching path(s): src/**/*
[nodemon] watching extensions: ts,json
[nodemon] starting `ts-node src/index.ts`
(node:65046) UnhandledPromiseRejectionWarning: DataTypeNotSupportedError: Data type "Object" in "UserSession.userAgent" is not supported by "postgres" database.
    at new DataTypeNotSupportedError (/Users/mandalarian/Github/accounts/src/error/DataTypeNotSupportedError.ts:9:9)
    at /Users/mandalarian/Github/accounts/src/metadata-builder/EntityMetadataValidator.ts:84:27
    at Array.forEach (<anonymous>)
    at EntityMetadataValidator.validate (/Users/mandalarian/Github/accounts/src/metadata-builder/EntityMetadataValidator.ts:81:36)
    at /Users/mandalarian/Github/accounts/src/metadata-builder/EntityMetadataValidator.ts:45:56
    at Array.forEach (<anonymous>)
    at EntityMetadataValidator.validateMany (/Users/mandalarian/Github/accounts/src/metadata-builder/EntityMetadataValidator.ts:45:25)
    at Connection.buildMetadatas (/Users/mandalarian/Github/accounts/src/connection/Connection.ts:522:33)
    at Connection.<anonymous> (/Users/mandalarian/Github/accounts/src/connection/Connection.ts:191:18)
    at step (/Users/mandalarian/Github/accounts/node_modules/tslib/tslib.js:139:27)
(node:65046) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
[...]

PR #994, opened 2020-06-03
fix(typeorm): add explicit types to userSession and IP decorator

theolampert:
This attempts to address the following issue when running the example repo for postgres + GraphQL (Close https://github.com/accounts-js/accounts/issues/991).

TypeORM has difficulty inferring underlying types from nullable types and will cast something like `string | null` into `object`. This PR adds an explicit type to typeORM's decorator.


pradel:
Thanks @theolampert!

flashtheman:
This fix seems not to be included in a release yet. I still face the issue with version 0.29.

PR #1000, opened 2020-06-09
fix: typeorm example fixes

darkbasic:
Make typeorm server working out of the box

### DIFF
--- a/examples/graphql-server-typeorm-postgres/.env
+++ b/examples/graphql-server-typeorm-postgres/.env
@@ -1,2 +1,2 @@
 DATABASE_URL=postgres://postgres@localhost:5432/postgres
-ACCOUNTS_SECRET=Change_me_please
+ACCOUNTS_SECRET=very-secret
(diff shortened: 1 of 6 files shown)


agustif:
I was thinking about dropping docker for the postgres db, and leave itto the user, could bee a nicee addo-on and be documented, but at least not have it by default on with start-servicees

darkbasic:
@agustif I completely agree and I've just made it optional

agustif:
@darkbasic Would you mind adding some details about both options and how to run them to the docs about postgres/typeorm?

darkbasic:
@agustif I will give it a try but if I recall correctly documentation doesn't build anymore in current git master

pradel:
@darkbasic the documentation should work on master, do you remember what was the issue you faced? 🤔 

For this example documentation, it should probably just be added in the README no?

darkbasic:
@pradel the readme was already correct in that regard because it assumed that docker wasn't mandatory (which wasn't true): I simply made that assumption stand.

Anyway I rephrased it to better explain the alternatives.

Regarding the docs, I've retried to build them and it seems to work now, so no clue what could have been the problem before.

pradel:
@darkbasic ok thanks! Can you please resolve the conflicts so I can merge? :)

darkbasic:
@pradel I've rebased it. should merge fine now

PR #1003, opened 2020-06-16
feat: add MikroOrm database adapter

darkbasic:
This will soon be updated to the upcoming MikroORM v4, but in the meantime it will allow me to integrate it into existing projects and improve the design. Right now it only supports PostgreSQL like the TypeORM one, but I see no issues adding support for other databases once v4 gets released.


darkbasic:
I've completely reworked the database adapter in order to easily make it extendable without having to re-implement Entities or rewrite any logic: it's just a matter of extending the base classes.
To achieve so the metadata is now stored in EntitySchemas and each of them is attached to a regular class (with no decorators because it doesn't have to store the metadata). Since the schema is generated by factories you can easily provide any custom class to extend the existing ones and everything will be handled gracefully. The example has been updated to show how to extend both User and Email, while retaining the existing Session and Service Entities.

At this point I could use some help to find out what's the easiest way to provide additional parameters to the DatabaseInterface methods. For example I extended User to hold additional props like name and surname, but I still need to provide those to createUser in order to store them when the account gets created.

darkbasic:
I wasn't fully satisfied with the previous implementation so I did yet another refactor: now the entity classes are decoupled from their interfaces and this allows me to use factories to generate their constructors. Since `User` embeds all the logic to create new `Email` and `Service` instances directly within its constructor, this allows me to hardcode the custom Email and Service constructors instead of having to provide them through args every time. The User factory also accepts two additional functions which will extract any additional parameters which may be required to create the aforementioned Email and Service instances.

At this point I'm happy enough to start integrating it with my existing application, but I still need to figure out how to provide additional parameters to the DatabaseInterface methods. I extended the `CreateUserInput` type (https://github.com/darkbasic/accounts/blob/mikro-orm/examples/graphql-server-mikro-orm-postgres/src/index.ts#L37) to provide the additional parameters and so far so good. The problem is that those parameters aren't passed to the `createUser` method (https://github.com/darkbasic/accounts/blob/mikro-orm/packages/database-mikro-orm/src/mikro-orm.ts#L108) of my custom implementation of `DatabaseInterface`. This is the same exact approach used by the TypeORM database adapter, so at this point I can only guess that it should be broken as well. What's the correct way to address this?

darkbasic:
The typeorm adapter was indeed broken in this regard: looking at the code it seems you need to pass a custom validateNewUser function in order for the additional field to not be filtered out.

pradel:
@darkbasic I am not familiar to MikroORM so I can't really review the code, but a couple of comments regarding the pr itself:
- would be great to have some documentation showing how to use it on the website.
- "Right now it only supports PostgreSQL", from the MikroORM docs I understand that even with v3 it should be possible to make it work with other databases. Would be really cool to have it working directly, do you think that would be possible?

Thanks for the work on this adapter :D

darkbasic:
> would be great to have some documentation showing how to use it on the website

Sure, I'll do it once it will be ready to get merged.

> "Right now it only supports PostgreSQL", from the MikroORM docs I understand that even with v3 it should be possible to make it work with other databases. Would be really cool to have it working directly, do you think that would be possible?

Should be possible, but I'm still not sure if I want the v3 version to be merged, because v4 is just around the corner and I plan to upgrade all my production apps as soon as it gets released. I think it will depend on the v4 timing compared to how much time it will require to finish this PR.

Right now I've just finished integrating it with one of my apps and while doing so I've decided that I wanted to rework the API to allow extending entities while retaining the original name. I also made several other changes and I'm not happy with it and how it integrates into existing projects: despite ORMs not usually being firendly in that regard it's just a matter of switching the graphql-module used for authentication.

@pradel Unfortunately now the priority is to tackle the biggest design flaw, one which I knew of since the beginning but purposely left as last: Mikro-ORM's Identity Map pattern (https://mikro-orm.io/docs/identity-map/).
Right now I'm using a shared `EntityManager` instance across all connections (https://github.com/darkbasic/accounts/blob/mikro-orm/packages/database-mikro-orm/src/mikro-orm.ts#L47), but that's wrong because of the Identity Map. Instead I need to fork it on each network request (or websocket message, depending on the transport). Considering the Accounts.js API is somehow transport agnostic, how would you suggest to do so?
Ideally, since I already have a graphl-modules provider forking it on each network request, it would be even better to find a way to reuse it so the identiy map is shared with the rest of the application on each network request:
```ts
providers: ({config: {orm}}) => [{
  provide: EntityManager,
  scope: ProviderScope.Session,
  useFactory: () => orm.em.fork(),
}],
```
```ts
export class SomeProvider {
  constructor(private em: EntityManager) {
```

darkbasic:
Maybe we could make `DatabaseInterface` (https://github.com/accounts-js/accounts/blob/master/packages/server/src/accounts-server.ts#L59) an `Injectable` and let the implementation decide which `ProviderScope` should be used? Basically anything else can still be a singleton (`ProviderScope.Application`) but it will allow db adapters like Mikro-ORM to use a session scope instead (I think Graphl-Modules doesn't support a per ws message scope, that should be implemented of course).
Inside my `DatabaseInterface` implementation I could use hooks to fork the em or simply depend on another provider for EntityManager in order to share it across the whole application.
This kind of change will affect the whole framework so this is something we need to discuss before I start looking into it.

pradel:
@darkbasic I am not sure I understood everything from your last message sorry :D 
Do you have an example I can look into?

darkbasic:
@pradel here it is: https://github.com/accounts-js/accounts/pull/1014

PR #1004, opened 2020-06-20
feat(graphql-api): custom typings for authenticated-resolver

TeNNoX:
I'm using it with nexus.js.org and it would break the types because it says 'any' for all of them.

### DIFF
--- a/packages/graphql-api/src/utils/authenticated-resolver.ts
+++ b/packages/graphql-api/src/utils/authenticated-resolver.ts
@@ -1,8 +1,8 @@
-export const authenticated = (func: any) => async (
-  root: any,
-  args: any,
-  context: any,
-  info: any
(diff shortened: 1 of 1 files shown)


packages/graphql-api/src/utils/authenticated-resolver.ts
```diff
@@ -1,8 +1,8 @@
-export const authenticated = (func: any) => async (
-  root: any,
-  args: any,
-  context: any,
-  info: any
+export const authenticated = <R, A, C, I, T>(func: (root: R, args: A, context: C, info: I) => T | Promise<T>) => async (
```
pradel, line 1: when using generics in the codebase we try to be more verbose so it's easier to understand, for example, `root` will be `CustomRoot` instead of `R` and `args` will be `CustomArgs` instead of `A`.

TeNNoX:
@pradel Does it still look acceptable? :sweat_smile:

PR #1014, opened 2020-07-07
Accounts overhaul

darkbasic:
Currently accounts-js' main GraphQL Module accepts instances of `AuthenticationServices` and `DatabaseInterfaces` through its config.
While this could be fine for the existing database adapters, some need to be instantiated on each network request. The Mikro ORM db adapter I'm currently contributing (https://github.com/accounts-js/accounts/pull/1003) is such an example. Mikro-ORM relies on the Identity Map pattern (https://mikro-orm.io/docs/identity-map/), which means that you will need to fork its `EntityManager` on each network request in order for its cache to pertain a specific user. Because of the way accounts-js is built that means we need to instantiate the `DatabaseInterface` (in this case `AccountsMikroOrm`) and the `AuthenticationService` (`AccountsPassword` or `AccountsOauth`) as well as the `AccountsServer` on each network request.
In order to do so the GraphQL Module cannot accept instances in its config: it needs to generate its own.
Initially I felt like there was some compelling reason in order to pass instances in the first place, so I started working on a new API based on hash maps (https://github.com/darkbasic/accounts/blob/hash-map/packages/graphql-api/src/modules/accounts/index.ts) where the key is the constructor and the value are the arguments, but as soon as I started to dig deeper into the codebase the more I felt like it was completely unnecessary: such an approach could be useful if you need to pass your custom `AuthenticationService` which lives outside of the accounts-js codebase, but that is not going to be working anyway because you will also need an additional GraphQL Module to implement the resolvers (I was initially mislead by the fact that `AccountsOauth` does not have one, but as I discovered that's simply because it still doesn't work with the GraphQL transport).
So the new approach is more radical and we simply pass the constructor parameters of the `DatabaseInterface` and the `AuthenticationServices`, while the GraphQL Module will use its own constructor to instantiate them. You can see an example here: https://github.com/darkbasic/accounts/blob/accounts-overhaul/examples/graphql-server-mikro-orm-postgres/src/index.ts#L23
In addition you can also pass an optional parameter to set the scope of the GraphQL Module providers (defaults to singleton). For those unfamiliar with GraphQL Modules the scope is basically the "policy" which dictates when it needs to create new instances: once per app, once per network request, etc.
The redisegned API works well, with a few limitations:
- hooks are broken
- accounts-boost needs to fixed, didn't bother to do so yet
- all examples except for the mikro-orm one need to be updated (trivial to do)

While working on the code I noticed that the design of the GraphQL Modules is very flawed because there are circular dependencies everywhere and modules like `AccountsPasswordModule` have dependencies on providers like `AccountsServer` provided by modules which they don't depend on. The only reason why this worked at all is because of the "go through each service and manually set the DatabaseInterface and AccountServer instances" thing: https://github.com/accounts-js/accounts/blob/master/packages/server/src/accounts-server.ts#L86
In order to alleviate the problem I moved `AccountsServer`, the `DatabaseInterface` and the `AuthenticationServices` into the `CoreAccountsModule`, so now `AccountsPasswordModule` can depend on it. Unfortunately, since `AccountsServer` depend on the `AuthenticationServices`, the `CoreAccountsModule` need to be in charge to instantiate `AccountsPassword` and `AccountsOauth` (while ideally their respective GraphQL Modules should be in charge of doing so).

It's still far from being pretty, but it's not worse than before. I would love to get some feedback and possibly some ideas for a better, more radical re-design.

Additional considerations:

- Currently this is how I fork the `EntityManager`: https://github.com/darkbasic/accounts/blob/accounts-overhaul/packages/database-mikro-orm/src/mikro-orm.ts#L54
While this works, that means that the instance won't be shared across the whole application, but only within accounts-js.
If you use GraphQL Module in your application you should create your own provider for `EntityManager` and `AccountsModule` should depend on it:
```ts
providers: ({config: {orm}}) => [
  {
    provide: EntityManager,
    scope: ProviderScope.Session,
    useFactory: () => orm.em.fork(),
  },
```
In order to depend on a provider you will need to depend on the module which provides it: I still have to experiment if it's possible to provide such module dependency at runtime through the config, but it would be nice to.

- While we already have `ProviderScope.Session` which allows to instantiate Providers on each network request, we still have to deal with websockets where we need to fork `EntityManager` on each **message**. A new scope should be created in GraphQL Modules, see https://github.com/Urigo/graphql-modules/issues/1218

@pradel @Urigo @dotansimha @ardatan

### DIFF
--- a/examples/graphql-server-mikro-orm-postgres/README.md
+++ b/examples/graphql-server-mikro-orm-postgres/README.md
@@ -1,0 +1,65 @@
+# mikro-orm-server-typescript
+
+This example demonstrate how to use [accounts-js](https://github.com/accounts-js/accounts) with PostgreSQL thanks to the [@accounts/mikro-orm](https://www.npmjs.com/package/@accounts/mikro-orm) plugin built by (https://github.com/darkbasic/)[@darkbasic].
+
+## Setup example
+
+In order to be able to run this example on your machine you first need to do the following steps:
+
+- Clone the repository `git clone git@github.com:accounts-js/accounts.git`
+- Install project dependencies: `yarn`
+- Link together all the packages: `yarn setup`
+- Compile the packages `yarn compile`
+- Go to the example folder `cd examples/graphql-server-mikro-orm-postgres`
+
+## Prerequisites
+
+If you're on a mac, you will probably already have postgres installed as a brew service.
+just edit the .env file and use your current postgres data
+
+You will need a PostgreSQL server to run this package. If you don't have a PostgreSQL server running already, and you have Docker & Docker Compose, you can do
+
+```bash
+docker-compose up -d
+```
+
+to start a new one.
+
+## Getting Started
+
+Start the app.
+
+Visit <http://localhost:4000/>
+
+```bash
+yarn start
+```
+
+-> [Start the client side](../react-graphql-typescript).
+
+```graphql
+mutation CreateUser {
+  createUser(
+    user: {
+      email: "john.does@john.com"
+      password: "1234567"
+      profile: { firstName: "John", lastName: "Doe" }
+    }
+  )
+}
+
+mutation Auth {
+  authenticate(
+    serviceName: "password"
+    params: { password: "1234567", user: { email: "john.does@john.com" } }
+  ) {
+    tokens {
+      accessToken
+    }
+  }
+}
+
+query Test {
+  privateField
+}
+```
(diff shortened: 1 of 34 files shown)


darkbasic:
Forgot to say that we also need to dynamically require (`createRequire`?) `AuthenticationServices`/`DatabaseInterfaces` because otherwise we would have to depend on them all.

PR #1017, opened 2020-07-11
docs: use gender neutral language in documentation

apaleslimghost:
the users your documentation refers to are not all people that would be referred to as 'him'. using gender neutral language is welcoming to more potential users of AccountsJS


pradel:
Thanks @apaleslimghost!

PR #1056, opened 2020-09-15
docs: fix an issue in an example of the README for graphql-api

JoJ123:
authenticated was just working with the proposed changes


pradel:
thanks @JoJ123 !

PR #1058, opened 2020-10-17
add Token separator option in accounts-client

agustif:
Using cookies for my future nextjs example renders tokens with an ugly 3%A as a seeparator, since cookies cant rendeer ':'

This is a fairly atomical change. hope it can get mergeed and doesent break anythhingg!!


agustif:
uh oh I added eveerythhing, closing for now will reopen llater

PR #1059, opened 2020-10-17
feat(accounts-client): Add option to customize token sepearator character

agustif:
When using cookies ':' renders as '%3A' as cookies only can handle some characters/encoding.

This lets me change the separator

### DIFF
--- a/packages/client/src/accounts-client.ts
+++ b/packages/client/src/accounts-client.ts
@@ -103,7 +103,7 @@ export class AccountsClient {
    * Refresh the user session
    * If the tokens have expired try to refresh them
    */
-  public async refreshSession(force: boolean = false): Promise<Tokens | null> {
+  public async refreshSession(force = false): Promise<Tokens | null> {
     const tokens = await this.getTokens();
     if (tokens) {
       try {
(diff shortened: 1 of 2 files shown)


pradel:
@agustif thanks for the pr! Looks like there are a lot of unrelated changes, can you please cleanup the pr to include only the intended changes? :)

I am wondering if it would be better to actually allow a custom function to change the key, that way you could really customise the key as you want. How does that sound?

agustif:
```
lerna ERR! yarn run compile exited 1 in '@accounts/client'
lerna ERR! yarn run compile stdout:
src/accounts-client.ts:98:35 - error TS2339: Property 'tokens' does not exist on type 'AuthenticationResult'.
  Property 'tokens' does not exist on type 'MultiFactorResult'.

98     await this.setTokens(response.tokens);
                                     ~~~~~~
 

Found 1 error.

```

Not sure why this is happenning with all combinations of using Promise<LoginResult | AuthenticationResult | MultifactorResult>

```
  /**
   * Login the user with a specific service
   */
  public async loginWithService(
    service: string,
    credentials: { [key: string]: any }
  ): Promise<AuthenticationResult> {
    const response = await this.transport.loginWithService(service, credentials);
    console.log(response);
    await this.setTokens(response.tokens);
    return response;
  }```

> @agustif thanks for the pr! Looks like there are a lot of unrelated changes, can you please cleanup the pr to include only the intended changes? :)
> 
> I am wondering if it would be better to actually allow a custom function to change the key, that way you could really customise the key as you want. How does that sound?

Yeah it's weird like I wasnt pushing from latest or something so MFA stuff wasnt there, cllosing this will reopen a new clean one

PR #1060, opened 2020-10-18
feat(accounts-client): Add token storage separator option to Client

agustif:
Reopened atomic-change PR.

Sorry for previous mess

Woulld use like this
```
const accountsClient = new AccountsClient({
  tokenStorage: cookieStorage,
  tokenStoragePrefix:"MyAccounts",
  tokenStorageSeparator: "-",
  },  
  accountsGraphQL
);
```


pradel:
@agustif I am wondering if it would be better to actually allow a custom function to change the key, that way you could really customise the key as you want.
We could deprecate the `tokenStoragePrefix` option and remove it in a future major version.
How does that sound?

agustif:
> @agustif I am wondering if it would be better to actually allow a custom function to change the key, that way you could really customise the key as you want.
> We could deprecate the `tokenStoragePrefix` option and remove it in a future major version.
> How does that sound?

Sounds fine for me, but not sure how to do it myself, can we close this and you will add it later?

pradel:
@agustif sorry I did some mistake while rebasing the pr 🤦 I pushed your change directly to master :) 
Thanks for the pr!

agustif:
> @agustif sorry I did some mistake while rebasing the pr 🤦 I pushed your change directly to master :)
> Thanks for the pr!

Yeah don't sweat about it bro! I don't even know how to rebase lol! Glad it landed onto master anyway! =)

PR #1085, opened 2021-01-05
fix(mongo): put mongodb types in dependencies

kiancross:
The `@accounts/mongo` package types depend on the `mongodb` types. Doesn't work with Yarn 2 PnP at the moment.

### DIFF
--- a/packages/database-mongo-password/package.json
+++ b/packages/database-mongo-password/package.json
@@ -36,7 +36,7 @@
   },
   "devDependencies": {
     "@types/jest": "26.0.15",
-    "@types/mongodb": "^3.5.33",
+    "@types/mongodb": "3.6.6",
     "@types/node": "14.14.9",
     "jest": "26.6.3",
     "ts-jest": "26.5.0"
--- a/packages/database-mongo-password/package.json
+++ b/packages/database-mongo-password/package.json
@@ -33,7 +33,8 @@
   },
   "devDependencies": {
     "@types/jest": "26.0.15",
-    "@types/mongodb": "3.5.33",
+    "@types/mongodb": "^3.5.33",
+
     "@types/node": "14.14.9",
     "jest": "26.6.3",
     "ts-jest": "26.4.0"
--- a/packages/database-mongo-password/package.json
+++ b/packages/database-mongo-password/package.json
@@ -37,7 +37,6 @@
   "devDependencies": {
     "@types/jest": "26.0.15",
     "@types/mongodb": "^3.5.33",
-
     "@types/node": "14.14.9",
     "jest": "26.6.3",
     "ts-jest": "26.5.0"
--- a/packages/database-mongo-sessions/package.json
+++ b/packages/database-mongo-sessions/package.json
@@ -36,7 +36,7 @@
   },
   "devDependencies": {
     "@types/jest": "26.0.15",
-    "@types/mongodb": "^3.5.33",
+    "@types/mongodb": "3.6.6",
     "@types/node": "14.14.9",
     "jest": "26.6.3",
     "ts-jest": "26.5.0"
--- a/packages/database-mongo-sessions/package.json
+++ b/packages/database-mongo-sessions/package.json
@@ -33,7 +33,7 @@
   },
   "devDependencies": {
     "@types/jest": "26.0.15",
-    "@types/mongodb": "3.5.33",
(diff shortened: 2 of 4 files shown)


packages/database-mongo/package.json
```diff
@@ -32,13 +32,13 @@
     "@accounts/types": "^0.30.0",
     "@accounts/mongo-password": "^0.30.0",
     "@accounts/mongo-sessions": "^0.30.0",
+    "@types/mongodb": "3.5.33",
```
pradel, line 35: We don't add the types in the dependencies to not impact non-typescript projects.
What is the error with yarn pnp?

kiancross: `IndexOptions`, which is exported by `mongodb`, is used in the signature of one of the public interfaces:

https://github.com/accounts-js/accounts/blob/3c2264db9fa121677201c85730625ee4f27ba54c/packages/database-mongo/src/mongo.ts#L63

Therefore `IndexOptions` is used in the generated type definitions.

Yarn PnP does not allow hoisting (as it is ambiguous) and so an error is thrown when the type definitions try to import `mongodb` (in actual fact, it is only `@types/mongodb` that are needed). This can be resolved by enabling `skipLibCheck`, however this should not be a requirement to use this package.

> We don't add the types in the dependencies to not impact non-typescript projects.

I think in most cases this is ok, as they are not needed at compile time for other packages. However there are a small number of cases (such as this), where they are. Additionally, many packages ship with their own type definitions anyway, so I'm not sure there is much to gain for non-typescript users by not including `@types/mongodb`.

pradel: Ok can we just unpin the version then? `"@types/mongodb": "^3.5.33",`

kiancross: Done

kiancross:
Looks like tests are failing after unpinning `@types/mongodb`? Guess the code doesn't support a newer version?

kiancross:
Ok fixed the other packages too. The lock file changed significantly too?

kiancross:
@pradel Any idea why the test is failing now?

packages/database-mongo-password/package.json
```diff
@@ -36,7 +36,8 @@
   },
   "devDependencies": {
     "@types/jest": "26.0.15",
-    "@types/mongodb": "3.5.33",
+    "@types/mongodb": "^3.5.33",
+
```
kiancross, line 40:
```suggestion
    "@types/mongodb": "3.5.33",
```

packages/database-mongo/package.json
```diff
@@ -29,16 +29,16 @@
   "author": "Leo Pradel",
   "license": "MIT",
   "dependencies": {
+    "@accounts/types": "^0.31.1",
     "@accounts/mongo-password": "^0.31.1",
     "@accounts/mongo-sessions": "^0.31.1",
```
kiancross, line 34:
```suggestion
    "@accounts/mongo-password": "^0.31.1",
    "@accounts/mongo-sessions": "^0.31.1",
    "@accounts/types": "^0.31.1",
```

packages/database-mongo-password/package.json
kiancross, line 40:
```suggestion
```

kiancross:
@pradel I think the versions in this PR are sorted now.

packages/database-mongo-password/package.json
```diff
@@ -36,7 +36,7 @@
   },
   "devDependencies": {
     "@types/jest": "26.0.15",
-    "@types/mongodb": "3.5.33",
+    "@types/mongodb": "^3.5.33",
```
pradel, line 39: Can you pin this one to 3.6.6 (latest npm tag)?

packages/database-mongo-sessions/package.json
```diff
@@ -36,7 +36,7 @@
   },
   "devDependencies": {
     "@types/jest": "26.0.15",
-    "@types/mongodb": "3.5.33",
+    "@types/mongodb": "^3.5.33",
```
pradel, line 39: Can you pin this one to 3.6.6 (latest npm tag)

pradel (commented):
I think you forgot to push the yarn.lock file :)

kiancross:
@pradel Done those suggestions and pushed lock file :)

pradel:
Thanks!

PR #1086, opened 2021-01-05
fix(graphql-api): add graphql to peerDependencies

kiancross:
`graphql` is a peer dependency of `@graphql-tools/utils` and needed at runtime.

### DIFF
--- a/packages/graphql-api/package.json
+++ b/packages/graphql-api/package.json
@@ -42,6 +42,7 @@
   "dependencies": {
     "@graphql-tools/merge": "6.2.5",
     "@graphql-tools/utils": "7.0.2",
+    "graphql": "14.6.0",
     "request-ip": "2.1.3",
     "tslib": "2.0.0"
   },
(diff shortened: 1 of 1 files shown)


packages/graphql-api/package.json
```diff
@@ -42,6 +42,7 @@
   "dependencies": {
     "@graphql-tools/merge": "6.2.5",
     "@graphql-tools/utils": "7.0.2",
+    "graphql": "14.6.0",
```
pradel, line 45: Multiple graphql versions usually creates problem, I think it would be better to have it as a "peerDependencies".

kiancross: @pradel Done!

packages/graphql-api/package.json
```diff
@@ -60,7 +61,6 @@
     "@types/jest": "25.2.3",
     "@types/request-ip": "0.0.35",
     "concurrently": "5.2.0",
-    "graphql": "14.6.0",
```
pradel, line 63: as it is now in peerDependencies we also need to have it as a devDependency otherwise we won't be able to import in in dev mode, can you please revert this line?

kiancross: Ok done. Also added support for v15, like in #1088.

pradel (approved):

PR #1089, opened 2021-01-09
fix(password): add @accounts/types to dependencies

kiancross:
Not sure if this should be in `dependencies` or `peerDependencies`, like here:

https://github.com/accounts-js/accounts/blob/0abf18649eddeeb17152caa5e30fdbbfc2c5bf99/packages/graphql-api/package.json#L38

### DIFF
--- a/packages/password/package.json
+++ b/packages/password/package.json
@@ -23,7 +23,6 @@
   },
   "dependencies": {
     "@accounts/two-factor": "^0.30.0",
-    "@accounts/types": "^0.30.0",
     "bcryptjs": "^2.4.3",
     "lodash": "^4.17.15",
     "tslib": "2.0.0"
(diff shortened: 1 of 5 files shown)


pradel:
It should be a "peerDependency" like the graphql-api 👍

kiancross:
> It should be a "peerDependency" like the graphql-api

Ok done!

packages/password/package.json
```diff
@@ -29,7 +29,6 @@
   },
   "devDependencies": {
     "@accounts/server": "^0.30.0",
-    "@accounts/types": "^0.30.0",
```
pradel, line 32: can you please rever this line?

kiancross:
Looks like someone else has already submitted a PR that moved that to peer dependencies, as master has been updated.

Issue #1096, opened 2021-01-26
stop impersonation from the client does not invalidate session in the server


blessanabraham:
# Bug report

## Describe the bug

When stopping impersonation in the client, the impersonated session is not invalidated in the server. Instead, the client only restores the original tokens in the storage. This ends up leaving the impersonation sessions in the server as still valid.

## To Reproduce

* Start a simple accounts server
* Login as a user
* Impersonate another user. You should see 2 new sessions in the server's database
* Call `stopImpersonation` in the client
* The impersonated token is still valid in the server's database

## Expected behaviour

Impersonated token's valid field in the database should be false

## System information

- Version of accounts-js: 0.30.0

PR #1097, opened 2021-01-26
fix(client): invalidate session when stopping impersonation

blessanabraham:
The session created when impersonating another user is is not
invalidated when the client stops impersonation. This PR adds a call to
logout to properly invalidate the session before restoring the original
tokens from storage.

Fixes: #1096


pradel:
Thanks for the fix!

Issue #1108, opened 2021-02-05
TypeORM findUserByServiceId query is wrong and hence returns the wrong result


blessanabraham:
# Bug report

## Describe the bug

The query in typeorm `findUserByServiceId` is wrong and hence returns the wrong results. The offending code can be found [here](https://github.com/accounts-js/accounts/blob/0abf18649eddeeb17152caa5e30fdbbfc2c5bf99/packages/database-typeorm/src/typeorm.ts#L169-L173)

The query there should actually be
```
const service = await this.serviceRepository.findOne(
  {
    name: serviceName,
    serviceId,
  },
  { cache: this.options.cache }
);
```

or:
```
const service = await this.serviceRepository.findOne({
  where: {
    name: serviceName,
    serviceId,
  },
  cache: this.options.cache,
 });
```

## System information

- Version of accounts-js: [e.g. 0.31.1]

PR #1109, opened 2021-02-05
fix(typeorm): fix query in findUserByServiceId #1108

blessanabraham:
This fixes the find query in `findUserByServiceId`

Fixes: #1108

### DIFF
--- a/packages/database-typeorm/src/typeorm.ts
+++ b/packages/database-typeorm/src/typeorm.ts
@@ -167,10 +167,12 @@ export class AccountsTypeorm implements DatabaseInterface {
 
   public async findUserByServiceId(serviceName: string, serviceId: string): Promise<User | null> {
     const service = await this.serviceRepository.findOne({
-      name: serviceName,
-      serviceId,
+      where: {
+        name: serviceName,
+        serviceId,
+      },
       cache: this.options.cache,
-    } as any);
+    });
 
     if (service) {
       return this.findUserById(service.userId);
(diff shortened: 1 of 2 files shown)


packages/database-typeorm/src/typeorm.ts
```diff
@@ -166,11 +166,13 @@ export class AccountsTypeorm implements DatabaseInterface {
   }
 
   public async findUserByServiceId(serviceName: string, serviceId: string): Promise<User | null> {
-    const service = await this.serviceRepository.findOne({
-      name: serviceName,
-      serviceId,
-      cache: this.options.cache,
-    } as any);
+    const service = await this.serviceRepository.findOne(
```
pradel, line 169: Should this follow the same pattern than line 111? (with the where condition)
```
    const service = await this.serviceRepository.findOne({
      where: {
        name: 'email.verificationTokens',
        token,
      },
      cache: this.options.cache,
    });
```

blessanabraham: Well, both patterns are used in this file. Personally, I prefer the where clause pattern (like you suggested). I will change it

pradel:
@blessanabraham can you uncomment the tests merged in the other pr? From my understanding, they should work with the fix right?

blessanabraham:
Yes, it's running as I am writing this

pradel:
Thanks for the fix @blessanabraham!

PR #1111, opened 2021-02-05
fix(typeorm): make the services attribute consistent with mongodb

### LINKED ISSUE #1110: TwoFactor service fails with typeorm
# Bug report

## Describe the bug

When using TwoFactor with typeorm, it fails to get the `two-factor` service from the user object because the `services` attribute is an array instead of an object. [See the code here0(https://github.com/accounts-js/accounts/blob/v0.31.1/packages/two-factor/src/utils/user.ts#L8)

The code for the `services` attribute construction in typeorm is [here](https://github.com/accounts-js/accounts/blob/v0.31.1/packages/database-typeorm/src/entity/User.ts#L45-L51)

This is also inconsistent with the service attribute on the `User` in mongodb (which returns it as an object instead of an array).

## System information

- Version of accounts-js: [e.g. 0.31.1]

ghost:
* Updated the `getServices` method on the `User` entity to construct
objects for the top level fields
* Added special handling for `email.verificationTokens` and
`password.reset`
* Updated tests to match the mongodb tests

Fixes: #1110

### DIFF
--- a/packages/database-typeorm/__tests__/typeorm-password.ts
+++ b/packages/database-typeorm/__tests__/typeorm-password.ts
@@ -62,11 +62,9 @@ describe('TypeormServicePassword', () => {
           },
         ],
         services: {
-          password: [
-            {
-              bcrypt: 'toto',
-            },
-          ],
+          password: {
+            bcrypt: 'toto',
+          },
         },
         createdAt: expect.any(Date),
         updatedAt: expect.any(Date),
@@ -81,12 +79,10 @@ describe('TypeormServicePassword', () => {
       });
       const ret = await accountsTypeorm.findUserById(userId);
       expect(userId).toBeTruthy();
-      expect(ret!.services).toMatchObject({
-        password: [
-          {
-            bcrypt: 'toto',
-          },
-        ],
+      expect(ret!.services).toEqual({
+        password: {
+          bcrypt: 'toto',
+        },
       });
     });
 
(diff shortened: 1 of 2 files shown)


blessanabraham:
The fix is ugly because I had to add special handling for some hard coded fields, namely `email.verificationTokens` and `password.reset`.
I am open to suggestions if there is a better way to do this. The only idea I have in my head right now is that, these fields  should be in their own tables and I would recommend the same for mongodb as well, since (a document has a size limit of 32 MB)[https://docs.mongodb.com/manual/reference/limits/] with the default storage. And since the email verification tokens get appended, we do not have control over the limit.

pradel:
> The only idea I have in my head right now is, these fields should be in their own tables and I would recommend the same for mongodb as well, since a document has a size limit of 16 MB with the default storage.

Yes totally agree, for mongo this is already in progress https://github.com/accounts-js/accounts/pull/1081

pradel:
@blessanabraham can you add some tests with various services for this function? As I am not super familiar with this part of the code I am afraid of breaking stuff.
Also from my understanding, this is a breaking change regarding the user format returned right?

blessanabraham:
sure, will add some tests if they are missing.I had to modify the tests that I previously added. from mongodb to match the response from typeorm. After this change, the tests. from mongodb and typeorm match up. I tested and made sure that TwoFactor works now with typeorm and after this change.
But you are right, I think we are missing some integration tests.

blessanabraham:
Regarding breaking change, no, this change keeps it consistent with what mongodb returns and what the services expect the Usre entity to look like. typeorm was broken before this change.

pradel:
@blessanabraham if you have time to add the tests I can create a release after it's merged with the changes :)

blessanabraham:
Hey @pradel , really sorry for the delay on this. Work was a bit hectic. I'll get to it this week

darkbasic:
@blessanabraham could you please rebase against current master and add some tests so that we can get this in the upcoming 1.0?

PR #1121, opened 2021-04-11
chore: Update peer dependencies to the latest version

Akarshit:
The `peerDependencies` were a little outdated. Since the version is still 0...., the `^` is behaving like a `~`. So while installing, the install was not working because of broken dependency tree.

### DIFF
--- a/packages/graphql-api/package.json
+++ b/packages/graphql-api/package.json
@@ -33,9 +33,9 @@
   },
   "homepage": "https://github.com/js-accounts/graphql-api",
   "peerDependencies": {
-    "@accounts/password": "^0.30.0",
-    "@accounts/server": "^0.30.0",
-    "@accounts/types": "^0.30.0",
+    "@accounts/password": "^0.31.1",
+    "@accounts/server": "^0.31.1",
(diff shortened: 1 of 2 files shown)


Akarshit:
@pradel Hey! I know you must be busy, but it would be great if you can have a look at this one.

pradel:
Thanks!

Akarshit:
Thanks for merging, I hope we get the new version soon....

pradel:
@Akarshit that's my bad but actually, but next version will be 0.32, so as soon as I release it will break again the peerDependency update. If you can create another pr with this I can create a new release just after.

Akarshit:
@pradel Just to be clear, I should search the codebase for `0.31.1` and update them all to `0.32.0`, right?

pradel:
just for peerDependencies yes

Akarshit:
@pradel Maybe I don't understand this fully, but when you release the next version, would it be `0.32.0` for all the packages? If yes should I update all the `package.json`(both dev and peer) to `0.32.0`?

pradel:
Next version will be 0.32.0 for all packages

PR #1123, opened 2021-04-20
chore: updating peer dependencies to latest version

Akarshit:
This PR updates the dependencies so that they would be up to date with the next release.


Akarshit:
@pradel I hope this works now.

PR #1147, opened 2021-05-30
feat: graphql api with graphql-module v1

tedraykov:
Following the graphql-modules 0.x to 1. migration guide,
added configuration.
Types, Queries and Mutations are now configurable
through module configuration.


pradel:
@tedraykov thank you so much for the pr!
I think in order to merge and release the pr we will need to update the documentation and examples to be sure everything is still working :)

darkbasic:
I started reviewing this, but it looks like it needs some fundamental changes in order to work.
First it doesn't even export the deprecated-accounts class which mimics the v0 behaviour, which is used by boost and all the examples (if fact they don't even compile). Even if you do export it, it won't work because it's not feature complete.
Also this approach is wrong in my opinion, because v1 is a completely new paradigm which gets rid of concepts like per-module contexts and thus we need to switch back to old school context generators like in the pre-gqlmodules days.
I don't think it's worth trying to mimic the old v0 API, I personally would go all in with the new API.
The other big change in v1 is that it doesn't enforce dependencies as strictly: in fact you cannot declare dependencies at all, just share some services with the global flag. That means there could be some room to re-think how the code is laid out. I still need some time before sharing my thoughts on this.

pradel:
@darkbasic yeah that was my assumption when I started working on the migration, I just wanted to have something working as soon as possible so the idea was to mimic the old API. I do agree that it's probably not the best way to do for the v1 and rethink the API is definitely the thing to do if possible.
I guess the big question is regarding the timeline for this, if we go with the simple approach I guess it can be done in the next weeks. Rethinking the API will take some time and work for sure but better for the long term.

darkbasic:
@pradel I don't think rethinking the public API would be too difficult, on the other hand rethinking how modules are internally laid out could prove to be harder. I will give you a more definitive answer in the next few days and ping you on Slack if I have any questions.
In the meantime could you please give me push access to the `feature/graphql-modules-v1` branch so I can get it working and we can start reasoning on the actual code?

pradel:
@darkbasic how can I do this, didn't find the option to give you permission on that branch :(

darkbasic:
Nevermind, I started from scratch because it was easier to keep track of the changes. I will create a new PR where we can discuss the new API.

PR #1149, opened 2021-06-11
fix: broken accounts-boost examples since graphql-tools 6+

darkbasic:
I was checking the examples after porting them to graphql-modules v1 and I've noticed that some of them were broken... since a long time. I've fixed the accounts-boost microservice example to work with graphql-tools 6+.


pradel:
Thanks @darkbasic !

PR #1150, opened 2021-06-14
feat: magic-link; password-free login

larsivi:
This is an addon service to password and adds the ability to authenticate via a token. The token can be sent to you via email using the requestLoginToken endpoint, or by other means server side by generating a token and adding it using addLoginToken. Users can now be found with this token, using findUserByLoginToken.
Default expiry is 15 minutes.

Three new packages are added, magic-link and mongo-magic-link and client-magic-link.

There is a logical dependency on package that handles users (typically password which in any case is a good companion module), but it is not hard if you have a different one providing the same.

Example of use (serverside, accompanied by password):
```ts
// There is nothing specific in Database setup, except the need for the mongo-magic-link package

const accountsPassword = new AccountsPassword({});
const accountsMagicLink = new AccountsMagicLink({});
const accountsServer = new AccountsServer(
       {
            db: accountsDb,
            tokenSecret: "terrible secret"
        },
        {
            password: accountsPassword,
            magicLink: accountsMagicLink,
        }
);
 ```

Example of use clientside (using GraphQL):
```ts
const apolloClient = new ApolloClient({ ...apolloOptions });
const accountsGraphQL = new GraphQLClient({graphQLClient: apolloClient});
accountsClient = new AccountsClient({}, accountsGraphQL);
const accountsMagicLink = new AccountsClientMagicLink(accountsClient);

 // Requesting email
await accountsMagicLink.requestMagicLinkEmail(email);

// Then, in the login handler
const loginResult = await accountsMagicLink.login({ token });
```


packages/token/src/accounts-token.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  public isTokenExpired(tokenRecord: TokenRecord, expiryDate: number): boolean {
+    return Number(tokenRecord.when) + expiryDate < Date.now();
+  }
+
+  private async tokenAuthenticator(token: string): Promise<CustomUser> {
+    const foundUser: CustomUser | null = await this.db.findUserByLoginToken(token);
+
+    if (!foundUser) {
+      if (this.server.options.ambiguousErrorMessages) {
```
pradel, line 110: I am not sure this is really needed as we authenticate with the login token and not an email here

pradel (commented):
Thank you very much for the pr, this will be a great addition that a lot of people requested!

As a general note, what do you think about renaming `@accounts/token` to `@accounts/magic-link`? The name is pretty standard for these kinds of authentication services, so it will be clear what this package is doing.

Regarding the logic, the implementation sounds good, I didn't do a deep review for this one but left a few small comments for now.
I was not able to compile the packages locally, graphql-client is throwing an error.

Security questions:
- when requesting 2 emails, should both be active or just the last one sent? So when I ask for a new email it invalidates all the previous tokens, only the last email link is working.

pradel:
More comments, documentation and examples are still missing for this, is this something you would be up to do?

larsivi:
> More comments: documentation and examples are still missing for this, is this something you would be up to do?

Yes, but if possible in a later PR.

packages/client-token/src/client-token.ts
```diff
@@ -0,0 +1,28 @@
...
+   */
+  public async login(user: LoginUserTokenService): Promise<LoginResult> {
+    return this.client.loginWithService('token', user);
+  }
+
+  /**
+   * Request a new login token.
+   * @param {string} email - The email address to send a password reset link.
+   */
+  public requestLoginToken(email: string): Promise<void> {
```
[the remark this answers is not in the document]
larsivi: I expect us to need this for SMS, which is why I went for a more generic variant, but I see your point (and I think you said you didn't want SMS support in accounts).

website/docs/strategies/password.md
```diff
@@ -88,7 +88,7 @@ The password module come with two factor out of the box. You can customize it us
 Check all the options available [here](/docs/api/two-factor/interfaces/accountstwofactoroptions).
 
 ```javascript
-export const accountsPassword = new AccountsPassword({
+export const accountsPassword = new AccountsToken({
```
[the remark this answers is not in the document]
larsivi: You're right.

packages/token/src/accounts-token.ts
larsivi: Exactly what isn't needed? Maybe I am misunderstanding something.

larsivi:
> As a general note, what do you think about renaming `@accounts/token` to `@accounts/magic-link`? The name is pretty standard for these kinds of authentication services, so it will be clear what this package is doing.

Sounds good!

> Security questions:
> 
> * when requesting 2 emails, should both be active or just the last one sent? So when I ask for a new email it invalidates all the previous tokens, only the last email link is working.
> * after a successful login all the tokens should be invalidated

Both of these makes sense, but maybe they could/should be covered by options? Or tied to the reason, such that login tokens explicitly added through addLoginToken could be managed by other means.

packages/client-magic-link/__tests__/client-magic-link.ts
```diff
@@ -0,0 +1,44 @@
...
+  loginWithService: jest.fn(),
+};
+const accountsMagicLink = new AccountsClientMagicLink(mockedClient as any);
+
+const user = {
+  email: 'john@doe.com',
+  token: 'deadbeef',
+};
+
+describe('AccountsClientPassword', () => {
```
pradel, line 16:
```suggestion
describe('AccountsClientMagicLink', () => {
```

packages/client-magic-link/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "files": [
+    "src",
+    "lib"
+  ],
+  "jest": {
+    "preset": "ts-jest"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/accounts-js/accounts/tree/master/packages/client-token"
```
pradel, line 31:
```suggestion
    "url": "https://github.com/accounts-js/accounts/tree/master/packages/client-magic-link"
```

packages/client-magic-link/src/client-magic-link.ts
```diff
@@ -0,0 +1,28 @@
...
+  /**
+   * Log the user in with a token.
+   */
+  public async login(user: LoginUserMagicLinkService): Promise<LoginResult> {
+    return this.client.loginWithService('magic-link', user);
+  }
+
+  /**
+   * Request a new login token.
+   * @param {string} email - The email address to send a password reset link.
```
pradel, line 23:
```suggestion
   * @param {string} email - The email address to send a login link.
```

packages/database-manager/__tests__/database-manager.ts
```diff
@@ -243,4 +255,16 @@ describe('DatabaseManager', () => {
   it('removeAllResetPasswordTokens should be called on userStorage', () => {
     expect(databaseManager.removeAllResetPasswordTokens('userId')).toBe('userStorage');
   });
+
+  it('addLoginToken should be called on sessionStorage', () => {
```
pradel, line 259:
```suggestion
  it('addLoginToken should be called on userStorage', () => {
```

packages/database-mongo-magic-link/src/mongo-magic-link.ts
```diff
@@ -0,0 +1,154 @@
...
+    this.password = this.options.password;
+  }
+
+  /**
+   * Setup the mongo indexes needed for the token service.
+   * @param options Options passed to the mongo native `createIndex` method.
+   */
+  public async setupIndexes(options: Omit<IndexOptions, 'unique' | 'sparse'> = {}): Promise<void> {
+    // Token index used to verify the email address of a user
+    await this.userCollection.createIndex('services.token.loginTokens.token', {
```
pradel, line 97: should we change token to magicLik here?

website/docs/transports/graphql.md
```diff
@@ -201,7 +201,7 @@ By default accounts-js only allow 'username', 'email' and 'password' for the use
 ```javascript
 // server.js
 
-const accountsPassword = new AccountsPassword({
+const accountsPassword = new AccountsMagicLink({
```
pradel, line 204: please revert

packages/types/src/types/services/magic-link/database-interface.ts
```diff
@@ -0,0 +1,9 @@
+import { User } from '../../user';
+
+export interface DatabaseInterfaceServiceMagicLink<CustomUser extends User = User> {
+  findUserByLoginToken(token: string): Promise<CustomUser | null>;
+
+  addLoginToken(userId: string, email: string, token: string, reason: string): Promise<void>;
```
pradel, line 6: do we really need a reason here?

pradel (commented):
- can you provide an example of the user creation? My understanding is that a password would be required in the current state? I might be missing something obvious tho :D

- I guess you deleted the `yarn.lock` file and it was regenerated from scratch, this is creating some issues, could you please revert the change you made and update the one from master instead of recreating it?

packages/magic-link/src/accounts-magic-link.ts
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  public isTokenExpired(tokenRecord: TokenRecord, expiryDate: number): boolean {
+    return Number(tokenRecord.when) + expiryDate < Date.now();
+  }
+
+  private async magicLinkAuthenticator(token: string): Promise<CustomUser> {
+    const foundUser: CustomUser | null = await this.db.findUserByLoginToken(token);
+
+    if (!foundUser) {
```
pradel, line 115: if user is not found it means that token is invalid, better to throw an `incorrectToken` error

pradel: also, can we rename `incorrectToken ` to `invalidToken` for consistency with password service?

pradel: actually I was taking a look at the password service and we should throw `LoginTokenExpired `. That way, an attacker can't guess if the token is expired or does not exist.

packages/magic-link/src/accounts-magic-link.ts
```diff
@@ -0,0 +1,147 @@
...
+        this.options.errors.userNotFound,
+        RequestMagicLinkEmailErrors.UserNotFound
+      );
+    }
+
+    // Remove pre-existing login tokens on user
+    await this.db.removeAllLoginTokens(user.id);
+
+    const token = generateRandomToken();
+    await this.db.addLoginToken(user.id, email, token, 'magic-link-requested');
```
pradel, line 74: why do we need a reason here?

packages/database-mongo-magic-link/src/mongo-magic-link.ts
larsivi: The service name should definately change to magicLink - was that what you meant?

packages/types/src/types/services/magic-link/database-interface.ts
larsivi: No, I am removing it.

packages/magic-link/src/accounts-magic-link.ts
larsivi: I have removed the reason param

packages/magic-link/src/accounts-magic-link.ts
larsivi: No ambigiousErrors handling? Always LoginTokenExpired? Would that also apply to the token record check itself?

pradel: Yeah I think so, to be consistent with https://github.com/accounts-js/accounts/blob/master/packages/password/src/accounts-password.ts#L307

pradel:
@larsivi 

```
// There is nothing specific in Database setup, except the need for the mongo-magic-link package

const accountsPassword = new AccountsPassword({});
const accountsMagicLink = new AccountsMagicLink({});
const accountsServer = new AccountsServer(
       {
            db: accountsDb,
            tokenSecret: "terrible secret"
        },
        {
            password: accountsPassword,
            magicLink: accountsMagicLink,
        }
);

```

regarding this, it's a bit weird that you have to add the password package too, this means that all the mutations (or routes via REST) will be exposed and user creation via password will be allowed.
Would it be possible to just install the magic link one?

larsivi:
> regarding this, it's a bit weird that you have to add the password package too, this means that all the mutations (or routes via REST) will be exposed and user creation via password will be allowed.
> Would it be possible to just install the magic link one?

Well, with the last cleanup I think that should work. The example was mostly like that since I do use the password module for user creation (and also for alternative login method). I haven't tried without though.

pradel:
Okay, would be great to have the example so we are sure it's working that way. Will make it easier for users to get started with magic-link!

acomito:
you guys are gods

pradel:
Amazing work @larsivi thank you so much for this!

pradel:
Everything is published on npm, would be cool to get some documentation for other users!

larsivi:
Thank you! I'll look at writing some docs over the next few weeks.

PR #1194, opened 2021-11-25
fix: repo links in graphql-api package.json

zenweasel:
Currently when you go to the page on NPM for this package it points to the old archived version which is confusing when the code actually lives here


zenweasel:
Tried to follow the instructions for adding a changeset but they didn't work for me

pradel:
Thanks for the fix, I added the changeset :)

PR #1203, opened 2021-12-12
docs: Update to GraphQL API Documentation

mrcleanandfresh:
Within the last month I discovered this project and started implementing it in my own. I found the docs to be a little on the lean side, so I thought I would contribute to them.

I added an example of how to authenticate with a service using GraphQL. I used most of the information [from this code](https://github.com/accounts-js/accounts/blob/220566425755a7015569d8e518095701ff7122e2/packages/graphql-api/src/modules/accounts/schema/types.ts). I was not sure of the "official" `serviceName` for each one of them though. 

I would also like to add additional information for each of the services, but am not sure what content to write, or if it should be organized differently. So this is my _first draft_ so to say. Please offer any feedback that would be helpful to fill out these docs a little more.

### DIFF
--- a/.changeset/slimy-pugs-refuse.md
+++ b/.changeset/slimy-pugs-refuse.md
@@ -1,0 +1,5 @@
+---
+'@accounts/graphql-api': patch
+---
+
+Added additional examples for how to authenticate with @accounts/graphql-api
--- a/packages/graphql-api/README.md
+++ b/packages/graphql-api/README.md
@@ -197,3 +197,80 @@ input CreateUserProfileInput {
 ```
 
 The user will be saved in the db with the profile key set.
+
+## Example: Authenticate with a Service
+
+```graphql
+mutation Authenticate {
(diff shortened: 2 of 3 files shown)


pradel:
Improving the documentation is definitely needed!
The update looks amazing, maybe you could add the same to the website documentation too what do you think?

mrclean-oa:
>  maybe you could add the same to the website documentation too what do you think?

Absolutely! Can you point me in the right direction, so I can make that update? I didn't realize there were two places that needed updating. Anything I should know when updating the website docs?

pradel:
All the website docs are located there https://github.com/accounts-js/accounts/tree/master/website/docs
Nothing in particular, we are using docusaurus and the website is there https://www.accountsjs.com/

mrcleanandfresh:
I've never used docusaurus so let me know if this is resolved in f1483c0. Didn't know how it determined when to start a new page, since I noticed the graphql.md file had both server and client in it.

pradel:
Looks great like this!

Issue #1204, opened 2021-12-12
Mongo Sessions: `convertSessionIdToMongoObjectId` not honored when `idProvider` is present


mrcleanandfresh:
# Bug report

## Describe the bug
I have this configuration for `MongoDbDriver`:

```typescript
await mongoose.connect(opts.databaseUrl, {
  useNewUrlParser: true,
  useUnifiedTopology: true,
});

const userStorage = new MongoDbDriver(mongoose.connection, {
  convertUserIdToMongoObjectId: false,
  idProvider: generateUserId,
  caseSensitiveUserName: true,
  collectionName: 'changedCollectionName',
  dateProvider: (date) => (date ? date : new Date()),
});
```
Note that I have the `idProvider` included.

When I logout a user then a Session gets invalidated, which means it has to be looked up, but [this line of code](https://github.com/accounts-js/accounts/blob/220566425755a7015569d8e518095701ff7122e2/packages/database-mongo-sessions/src/mongo-sessions.ts#L170) is causing problems. When that line executes, `convertSessionIdToMongoObjectId` is `true`, which makes sense, but the value of `sessionId` **is not** a Mongo Object ID. So I'm assuming the inclusion of `idProvider` is causing it to ignore my configuration and applying that provider to both ids.

## To Reproduce

- Include an `idProvider` that prepends or appends a value to a Mongo Object Id or otherwise changes it like so:
```typescript
export function generateUserId(): string {
  return `somevalue|${new mongoose.Types.ObjectId(Date.now())}`;
}
```
- Once a session is created in the MongoDB, query it and you'll notice that the id is generated the same way regardless of the `convert*ToMongoObjectId` configuration.

## Expected behavior

I would expect if a `convert*ToMongoObjectId` is set to `true` then it would ignore any `idProvider` that is passed into the configuration, and generate a bonafide Mongo ObjectId.

## Screenshots

![badutilcall](https://user-images.githubusercontent.com/10137846/145721890-c3ec5583-762d-412b-8ac2-28421b4424a7.png)

## System information

-  "@accounts/database-manager": "^0.33.1"
-  "@accounts/graphql-api": "^0.33.1"
-  "@accounts/magic-link": "^0.1.1"
-  "@accounts/mongo": "^0.33.5"
-  "@accounts/password": "^0.32.1"
-  "@accounts/server": "^0.33.1"
-  "mongoose": "5.9.25"

## Additional context

[...]

PR #1205, opened 2021-12-12
fix: check `convertSessionIdToMongoObjectId` before using `idProvider`

mrcleanandfresh:
This fix is related to this bug report: https://github.com/accounts-js/accounts/issues/1204

The configuration option of `convertSessionIdToMongoObjectId` was not being checked when using the `idProvider` to create a new session.

### DIFF
--- a/.changeset/swift-panthers-play.md
+++ b/.changeset/swift-panthers-play.md
@@ -2,4 +2,8 @@
 '@accounts/mongo-sessions': minor
 ---
 
-Changed `idProvider` to `idSessionProvider` in mongo-sessions. If you are using `idProvider` for the creation of a custom identifier for sessions, then you will need to move that logic to a new `idSessionProvider` function in the configuration. Going forward, the `idProvider` will only be used for the creation of user identifiers. This change was made to give more granular control when creating custom identifiers for both sessions and users.
+Changed `idProvider` to `idSessionProvider` in `@mongo-sessions`. This change was made to give more granular control when creating custom identifiers for both sessions and users. The new option generates the _id for new Session objects. Going forward, the `idProvider` will only be used for the creation of user identifiers.
+
+Updated `AccountsMongoOptions` in `@database-mongo` to have the new `idSessionProvider`.
+
+**Migration:** If you are using `idProvider` for the creation of a custom identifier for sessions, then you will need to move that logic to the new `idSessionProvider` function in the configuration.
--- a/.changeset/swift-panthers-play.md
+++ b/.changeset/swift-panthers-play.md
@@ -2,4 +2,4 @@
 '@accounts/mongo-sessions': minor
 ---
 
-Changed idProvider to idSessionProvider in mongo-sessions
+Changed `idProvider` to `idSessionProvider` in mongo-sessions. If you are using `idProvider` for the creation of a custom identifier for sessions, then you will need to move that logic to a new `idSessionProvider` function in the configuration. Going forward, the `idProvider` will only be used for the creation of user identifiers. This change was made to give more granular control when creating custom identifiers for both sessions and users.
--- a/.changeset/swift-panthers-play.md
+++ b/.changeset/swift-panthers-play.md
@@ -1,0 +1,5 @@
+---
+'@accounts/mongo-sessions': minor
+---
+
+Changed idProvider to idSessionProvider in mongo-sessions
--- a/.changeset/swift-panthers-play.md
+++ b/.changeset/swift-panthers-play.md
@@ -5,6 +5,6 @@
 
 Changed `idProvider` to `idSessionProvider` in `@mongo-sessions`. This change was made to give more granular control when creating custom identifiers for both sessions and users. The new option generates the _id for new Session objects. Going forward, the `idProvider` will only be used for the creation of user identifiers.
 
-Updated `AccountsMongoOptions` in `@database-mongo` to have the new `idSessionProvider`.
+Updated `AccountsMongoOptions` in `@mongo` to have the new `idSessionProvider`.
 
 **Migration:** If you are using `idProvider` for the creation of a custom identifier for sessions, then you will need to move that logic to the new `idSessionProvider` function in the configuration.
--- a/.changeset/swift-panthers-play.md
+++ b/.changeset/swift-panthers-play.md
@@ -1,5 +1,6 @@
 ---
 '@accounts/mongo-sessions': minor
+'@accounts/database-mongo': minor
 ---
 
 Changed `idProvider` to `idSessionProvider` in `@mongo-sessions`. This change was made to give more granular control when creating custom identifiers for both sessions and users. The new option generates the _id for new Session objects. Going forward, the `idProvider` will only be used for the creation of user identifiers.
--- a/.changeset/swift-panthers-play.md
+++ b/.changeset/swift-panthers-play.md
@@ -1,6 +1,6 @@
 ---
 '@accounts/mongo-sessions': minor
-'@accounts/database-mongo': minor
+'@accounts/mongo': minor
 ---
 
 Changed `idProvider` to `idSessionProvider` in `@mongo-sessions`. This change was made to give more granular control when creating custom identifiers for both sessions and users. The new option generates the _id for new Session objects. Going forward, the `idProvider` will only be used for the creation of user identifiers.
--- a/packages/database-mongo-sessions/src/mongo-sessions.ts
+++ b/packages/database-mongo-sessions/src/mongo-sessions.ts
@@ -33,8 +33,7 @@ export interface MongoSessionsOptions {
    */
   idSessionProvider?: () => string | object;
   /**
-   * Function that generate the id for new objects.
-   * @deprecated - please use `idSessionProvider` instead.
+   * Function that generate the id for new User objects.
    */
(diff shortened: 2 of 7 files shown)


mrclean-oa:
> What do you think about introducing a new idSessionProvider instead?

I think that would be great! I think we have a couple of options:

1. **Use both the boolean and the provider functions** (recommended) - It seems redundant, but I think the `convertSessionIdToMongoObjectId` and `convertUserIdToMongoObjectId` booleans need to be consulted prior to creating a new user or a new session. This is in addition to checking whether the `id*Provider` function is present.
2. **Remove the booleans that configure the override** - it is confusing that they are configured in order to use a regular ObjectId, but are not consulted when the object id is being created. I don't recommend this option because of the check that is in place to convert the id to an ObjectId when querying.
3. Other options?

What are your thoughts?

mrcleanandfresh:
> What do you think about introducing a new idSessionProvider instead?

1. **Use both the boolean and the provider functions** (recommended) - It seems redundant, but I think the convertSessionIdToMongoObjectId and convertUserIdToMongoObjectId booleans need to be consulted prior to creating a new user or a new session. This is in addition to checking whether the id*Provider function is present. This provides more guarantees about the data that makes it into the Database.
2. **Remove the booleans that configure the override** - it is confusing that they are configured in order to use a regular ObjectId, but are not consulted when the object id is being created. I don't recommend this option because of the check that is in place to convert the id to an ObjectId when querying.
    Other options?

What are your thoughts?

pradel:
To avoid a maximum the breaking changes and impact the users when they will upgrade to the new version option 1 seems good :)

> but I think the convertSessionIdToMongoObjectId and convertUserIdToMongoObjectId booleans need to be consulted prior to creating a new user or a new session

I am just not sure about this, this would mean that idProvider is actually not used if the value is false right? That can be really confusing.

pradel:
The warning approach sounds like a good idea 👍

mrcleanandfresh:
> The warning approach sounds like a good idea 👍

The approach I'm thinking about will be very simple:
1. Add documentation about these two configuration variables (and likely all the others as well)
2. Add additional condition for where `idProvider` is used for the user, to check for `!convertUserIdToMongoObjectId`
3. For both add an if-statement in the constructor that checks. Example for `idProvider` and `convertUserIdToMongoObjectId`:
```typescript
if (typeof idProvider === 'function' && !convertUserIdToMongoObjectId) {
  console.warn(`You have set both "idProvider" and "convertUserIdToMongoObjectId = false" which will cause your "idProvider" to be ignored. 
In order to fix this warning change "convertUserIdToMongoObjectId" to "true" or remove your "idProvider" from the configuration.
`);
}
```
4. Add a test that does this and spies on `console.warn` to be called with `idProvider` and `convertUserIdToMongoObjectId`.

What do you think of that approach, anything I'm missing?

Do we have existing docs for the configuration options or will that be a new section? If it will be new how would it be best organized? Otherwise, can you point me to the location for where it will live?

Thanks for any feedback!

mrcleanandfresh:
I also couldn't help but notice that the [Mongo Options](https://www.accountsjs.com/docs/databases/mongo#options) contain configuration that assumes you will use Accounts Password in it, see `convertUserIdToMongoObjectId` and `idProvider`. Should that be separated out?

packages/database-mongo-sessions/src/mongo-sessions.ts
```diff
@@ -61,6 +69,23 @@ export class MongoSessions implements DatabaseInterfaceSessions {
       timestamps: { ...defaultOptions.timestamps, ...options.timestamps },
     };
 
+    if (typeof this.options.idProvider === 'function') {
```
pradel, line 72: The issue I see of this check is that currently this option is inherited from the @accounts/mongo package, see https://github.com/accounts-js/accounts/blob/master/packages/database-mongo/src/mongo.ts#L58. The user might use both `idProvider` and `idSessionProvider` and this condition would be triggered.
We should probably remove this totally.

.changeset/swift-panthers-play.md
```diff
@@ -0,0 +1,5 @@
+---
+'@accounts/mongo-sessions': minor
+---
+
+Changed idProvider to idSessionProvider in mongo-sessions
```
pradel, line 5: As this is a breaking change can we provide more info for the user to explain how to upgrade?

packages/database-mongo-sessions/src/mongo-sessions.ts
```diff
@@ -61,6 +69,23 @@ export class MongoSessions implements DatabaseInterfaceSessions {
...
+        `Deprecation Warning: The option "options.idProvider" is deprecated. 
+        Setting it to "options.idSessionProvider," please change your configuration, as "options.idProvider" will be removed in a future release.`
+      );
+    }
+
+    if (
+      typeof this.options.idSessionProvider === 'function' &&
+      this.options.convertSessionIdToMongoObjectId
+    ) {
+      console.warn(`You have set both "options.idSessionProvider" and "options.convertSessionIdToMongoObjectId = false" which will cause your "options.idSessionProvider" to be ignored.
```
pradel, line 84: I am a bit confused I think, shouldn't it be "options.convertSessionIdToMongoObjectId = true"?

mrcleanandfresh: Yes! Good catch. Typo. Addressed in b89b6cdf77c759f9dc2dc7de936becf805184259

.changeset/swift-panthers-play.md
mrcleanandfresh: That's an excellent point. This is addressed in b89b6cdf77c759f9dc2dc7de936becf805184259

packages/database-mongo-sessions/src/mongo-sessions.ts
mrcleanandfresh: I have removed this and fixed a failing test to use `idSessionProvider` since it was previously using `idProvider`. This is addressed in b89b6cdf77c759f9dc2dc7de936becf805184259.

mrcleanandfresh:
@pradel I see that the "Test Packages" Github workflow is failing. Since this is a breaking change, I feel like that is expected, but could use your guidance on how to proceed.

pradel:
The tests failing are coming from here, https://github.com/accounts-js/accounts/blob/master/packages/database-tests/src/index.ts
To run it on your computer, you can run `yarn test` in the mongo folder :)

pradel:
Yes, you might have to edit the mongo package to pass down the options for the tests to run properly :)

mrcleanandfresh:
@pradel tests are passing. Let me know once you've tested some things. I had to add a type definition to the `AccountsMongoOptions` interface, so just want to be sure I don't need to add the `idSessionProvider` to any other places to ensure no breakage with TypeScript.

mrcleanandfresh:
I get this error when I try to run all tests:

```bash
~/Code/forks/accounts$ pnpm run testonly

> @0.22.0 testonly /home/kevin/Code/forks/accounts
> lerna run testonly

lerna notice cli v4.0.0
lerna info Executing command in 23 packages: "pnpm run testonly"
lerna ERR! pnpm run testonly exited 1 in '@accounts/oauth-twitter'
lerna ERR! pnpm run testonly stdout:

> @accounts/oauth-twitter@0.32.1 testonly /home/kevin/Code/forks/accounts/packages/oauth-twitter
> jest

 ELIFECYCLE  Command failed with exit code 1.
lerna ERR! pnpm run testonly stderr:
FAIL __tests__/index.ts
  ● Test suite failed to run

    TypeError: Cannot read property 'bind' of undefined

      at Runtime._createJestObjectFor (../../node_modules/.pnpm/jest@27.3.1/node_modules/jest/node_modules/jest-runtime/build/index.js:2193:46)

Test Suites: 1 failed, 1 total
Tests:       0 total
Snapshots:   0 total
Time:        0.619 s
Ran all test suites.
lerna ERR! pnpm run testonly exited 1 in '@accounts/oauth-twitter'
lerna WARN complete Waiting for 3 child processes to exit. CTRL-C to exit immediately.
 ELIFECYCLE  Command failed with exit code 1.
```
Do you know what I'm missing, or should I not be trying to run the tests at this level?

pradel:
I never run the tests from the root of the monorepo so don't know about this error, you can run the testonly command in the subfolders too :)

.changeset/swift-panthers-play.md
```diff
@@ -0,0 +1,5 @@
+---
+'@accounts/mongo-sessions': minor
```
pradel, line 2: I think that this change should also update the mongo package as AccountsMongoOptions was changed

pradel:
> Also, please refer to a previous question about the strategy for updating the docs. If the latter is preferred, and you can offer clear enough directions on how I can find all combinations I can take care of this now. Otherwise, it might be good for you or someone else to tackle that in a separate PR.

We can leave this to a separate pr, let's merge and release the work you did first :)

.changeset/swift-panthers-play.md
mrcleanandfresh: Resolved in 710f357

pradel (approved):

mrcleanandfresh:
@pradel Absolutely, happy to contribute!

PR #1206, opened 2021-12-14
feat: setup turbomodule

pradel:
Just a test for now

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,5 +10,6 @@ packages/*/package-lock.json
 schema.json
 .DS_Store
 .eslintcache
+.turbo
 
 build/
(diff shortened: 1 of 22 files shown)


darkbasic:
Unfortunately turbo is not a viable option because they're not multiplatform anymore:
https://github.com/vercel/turbo/issues/5957

Lerna 7 isn't either: https://github.com/lerna/lerna/issues/3676

At the moment we're still using lerna 6 (so I am actually able to use it on ppc64le) but I've already migrated to yarn 4 which bundles a brand new task runner. I'm planning to use it for the most basic tasks like compiling and testing, like I did in mikro-orm: https://github.com/mikro-orm/mikro-orm/pull/4331
Everything else is for CI consumption only so we can stick with lerna for it.

PR #1180, opened 2021-10-11
fix: fix compatibility issue with MongoDB 4.X

rawb1:
Could be a good idea to bump mongo version ? I did not find any information on mongo version compatibility ...

Fixes #1174

PR #1209, opened 2022-01-14
fix: patch for issue in mongo driver 4.x.x

gezquinndesign:
I know some issues were addressed in #1180.  These are a couple of other issues that arise because in this version of the driver, the result of updateOne() and updateMany() is now different.  Keeping previous check in there for backwards compatibility.


pradel:
Can you add the changeset file to the pr? Lgtm!

gezquinndesign:
Added. :)

Issue #1219, opened 2022-05-14
`TypeError: Cannot read property 'nModified' of undefined` calling `MongoServicePassword.setPassword`


KeithGillette:
# Bug report

## Describe the bug

Attempting to integrate AccountsJS into a project using `mongodb@4.5.0` using `mongoose@6.3.2` and `@typegoose/typegoose@9.8.2`, we get the error `TypeError: Cannot read property 'nModified' of undefined` when calling `MongoServicePassword.setPassword`.

## To Reproduce

See description.

## Expected behavior

`MongoServicePassword.setPassword` will return a void Promise on success per the documentation or `throw new Error('User not found')` per the code.

## System information
```
    "@accounts/apollo-link": "0.32.1",
    "@accounts/client": "0.33.1",
    "@accounts/client-password": "0.32.2",
    "@accounts/graphql-api": "0.33.2",
    "@accounts/graphql-client": "0.33.1",
    "@accounts/magic-link": "0.1.1",
    "@accounts/mongo": "0.34.0",
    "@accounts/password": "0.32.1",
    "@accounts/server": "0.33.1",
    ...
    "@typegoose/typegoose": "9.8.1",
    ...
    "mongodb": "4.5.0",
    "mongoose": "6.3.2",
```

## Additional context

[The source code for `MongoServicePassword.setPassword` attempts to reference `ret.result.nModified`](https://github.com/accounts-js/accounts/blob/master/packages/database-mongo-password/src/mongo-password.ts#L351), as does `verifyEmail` & `setUsername`) but `addEmail` & `removeEmail` first test for `ret.modifiedCount`.

In my testing, the `ret` `const` directly contains the `updateOne` result of the shape:
```
{
   acknowledged: true,
   modifiedCount: 1,
   upsertedId: null,
   upsertedCount: 0,
   matchedCount: 1
}
```

I believe all `MongoServicePassword` methods should be updated to test for `ret.modifiedCount`.

pradel:
Indeed, pr welcome for the fix, looks like these methods were not updated 👍

pradel:
New version was published with a fix for this one.

PR #1220, opened 2022-05-18
fix: mongodb 5 compatability in all mongo password methods

abhishekk-raj:
Fix for issue #1219


pradel:
@abhishekk-raj could please add a changeset to the pr? Otherwise LGTM 🚀

Issue #1227, opened 2022-06-28
TwoFactor Bypass Security Hole


pozylon:
# Bug report

## Describe the bug

Two Factor accepts codes that are not valid, you can generate a secret and afterwards login with the TOTP code 123124, completely bypasses the second factor.

## To Reproduce

1. getTwoFactorSecret
2. twoFactorSet with correct code
3. login with your password and TOTP 123124

## Expected behavior

No login because code incorrect

## System information

- accounts-password 0.32.3
- two-factor 0.32.3

## Additional context

The problem has been fixed lately with a new version of the fork of @levminer here https://github.com/Levminer/speakeasy/issues/11

So a simple bump should solve it

PR #1228, opened 2022-06-28
fix: totp code validation

pozylon:


pozylon:
Fixes #1227

pradel:
@pozylon thanks for the pr!

A couple of changes are needed for the ci to pass so I can merge and release:
- do not change the `version` field, it will be changed automatically by the ci when we release the next version
- revert the change made to `packages/password/package.json`, same the ci will take care of it
- can you please add a changeset explaining the change and which packages needs to be bumped
- run `pnpm install` to update the lockfile of the repo and commit it

pradel:
@pozylon looks like with the upgrade, tests are failing

pozylon:
@pradel The new version of speakeasy started to throw a custom exception when the token length did not match, I guarded it now so tests of two-factor passed locally but let's see what the workflow says.

pradel:
@pozylon this looks like a breaking change to me, some people may be catching some errors to return specific error messages to users, could we rather fix the tests or make this specific error silent?

pozylon:
@pradel It's not breaking because version 1.3.1 of speakeasy did not throw any exception during token validation, https://github.com/Levminer/speakeasy/blob/1.3.1/main.js (before) vs
https://github.com/Levminer/speakeasy/blob/1.3.3/main.js (now)

pozylon:
@pradel So to outline the changes in speakeasy, it's 3 things:

1. throws now if token length mismatch, did not throw before
2. throws now if token is not a number, did not throw before
3. only returns the delta in verifyDelta when there is a code match, did also return a delta in "else" thus validating almost all codes as true

```javascript
exports.hotp.verifyDelta = hotpVerifyDelta = (options) => {
	let i

	// shadow options
	options = Object.create(options)

	// unpack options
	let token = String(options.token)
	const digits = parseInt(options.digits, 10) || 6
	const window = parseInt(options.window, 10) || 0
	const counter = parseInt(options.counter, 10) || 0

	// fail if token is not of correct length
	if (token.length !== digits) {
		return
	}

	// parse token to integer
	token = parseInt(token, 10)

	// fail if token is NA
	if (isNaN(token)) {
		return
	}

	// loop from C to C + W inclusive

	// this is a temporary fix becaouse wrong codes return a value too
	// FIX THIS LATER
	// eslint-disable-next-line no-unreachable-loop
	for (i = counter; i <= counter + window; ++i) {
		options.counter = i
		// domain-specific constant-time comparison for integer codes
		if (parseInt(exports.hotp(options), 10) === token) {
			// found a matching code, return delta
			return { delta: i - counter }
		} else {
			return { delta: i - counter }
		}
	}

	// no codes have matched
}
```

```javascript
exports.hotp.verifyDelta = hotpVerifyDelta = (options) => {
	let i

	// shadow options
	options = Object.create(options)

	// unpack options
	let token = String(options.token)
	const digits = parseInt(options.digits, 10) || 6
	const window = parseInt(options.window, 10) || 0
	const counter = parseInt(options.counter, 10) || 0

	// fail if token is not of correct length
	if (token.length !== digits) {
		throw new Error("@levminer/speakeasy - hotpVerifyDelta - Wrong toke length")
	}

	// parse token to integer
	token = parseInt(token, 10)

	// fail if token is NA
	if (isNaN(token)) {
		throw new Error("@levminer/speakeasy - hotpVerifyDelta - Token is not a number")
	}

	// loop from C to C + W inclusive
	for (i = counter; i <= counter + window; ++i) {
		options.counter = i
		// domain-specific constant-time comparison for integer codes
		if (parseInt(exports.hotp(options), 10) === token) {
			// found a matching code, return delta
			return { delta: i - counter }
		}
	}

	// no codes have matched
}
```

pradel:
Okay thanks for the explanation, let's go with it then!

Issue #1250, opened 2023-05-08
Mongo.createUser and Mongo.createSession on Meteor@2.5.1


awatson1978:
# Bug report

Getting the following error on Meteor@2.5.1, which uses Mongo 4.X:

```
W20230507-20:07:34.414(-5)? (STDERR) error TypeError: Cannot read property '0' of undefined
W20230507-20:07:34.414(-5)? (STDERR)     at Mongo.<anonymous> (/Users/awatson/Code/nomex/nomex-app/node_modules/@accounts/mongo/lib/mongo.js:454:54)
W20230507-20:07:34.414(-5)? (STDERR)     at step (/Users/awatson/Code/nomex/nomex-app/node_modules/@accounts/mongo/node_modules/tslib/tslib.js:140:27)
W20230507-20:07:34.414(-5)? (STDERR)     at Object.next (/Users/awatson/Code/nomex/nomex-app/node_modules/@accounts/mongo/node_modules/tslib/tslib.js:121:57)
W20230507-20:07:34.414(-5)? (STDERR)     at fulfilled (/Users/awatson/Code/nomex/nomex-app/node_modules/@accounts/mongo/node_modules/tslib/tslib.js:111:62)
W20230507-20:07:34.415(-5)? (STDERR)     at /Users/awatson/.meteor/packages/promise/.0.12.0.qcz8u8.npe48++os+web.browser+web.browser.legacy+web.cordova/npm/node_modules/meteor-promise/fiber_pool.js:43:40
W20230507-20:07:34.415(-5)? (STDERR)  => awaited here:
W20230507-20:07:34.415(-5)? (STDERR)     at Function.Promise.await (/Users/awatson/.meteor/packages/promise/.0.12.0.qcz8u8.npe48++os+web.browser+web.browser.legacy+web.cordova/npm/node_modules/meteor-promise/promise_server.js:56:12)
W20230507-20:07:34.415(-5)? (STDERR)     at server/AccountsServer.js:1009:24
W20230507-20:07:34.415(-5)? (STDERR)     at /Users/awatson/.meteor/packages/promise/.0.12.0.qcz8u8.npe48++os+web.browser+web.browser.legacy+web.cordova/npm/node_modules/meteor-promise/fiber_pool.js:43:40
W20230507-20:07:34.417(-5)? (STDERR) (node:5433) UnhandledPromiseRejectionWarning: TypeError: Cannot read property '0' of undefined
[...]

PR #1251, opened 2023-05-08
Bugfix on Mongo.createUser and Mongo.createSession on 4.X

awatson1978:
Bug fix for #1250

### DIFF
--- a/packages/database-mongo-sessions/src/mongo-sessions.ts
+++ b/packages/database-mongo-sessions/src/mongo-sessions.ts
@@ -124,7 +124,9 @@ export class MongoSessions implements DatabaseInterfaceSessions {
 
     const ret = await this.sessionCollection.insertOne(session);
     // keep ret.ops for compatibility with MongoDB 3.X, version 4.X uses insertedId
-    return ((ret.insertedId ?? ret.ops[0]._id) as ObjectID).toString();
+    // return ((ret.insertedId ?? ret.ops[0]._id) as ObjectID).toString();
+    return ((ret.insertedId ? ret.insertedId : ret.ops[0]._id) as ObjectID).toString();
+
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/database-mongo-sessions/src/mongo-sessions.ts
```diff
@@ -124,7 +124,9 @@ export class MongoSessions implements DatabaseInterfaceSessions {
 
     const ret = await this.sessionCollection.insertOne(session);
     // keep ret.ops for compatibility with MongoDB 3.X, version 4.X uses insertedId
-    return ((ret.insertedId ?? ret.ops[0]._id) as ObjectID).toString();
+    // return ((ret.insertedId ?? ret.ops[0]._id) as ObjectID).toString();
```
pradel, line 127: can you please remove the commented line?

awatson1978: Done.  [c669c93](https://github.com/accounts-js/accounts/pull/1251/commits/c669c93019aeb9385b596c6388d47d5ec27f0c86)

awatson1978:
Okay, I added a changeset.  First time I've ever done that, so double check it to make sure it looks right.

package.json
```diff
@@ -77,5 +77,8 @@
     "type": "opencollective",
     "url": "https://opencollective.com/accounts-js",
     "logo": "https://opencollective.com/accounts-js/logo.txt"
+  },
```
pradel, line 80: can you please delete changeset here? The cli is already set in the dev deps :)

awatson1978: Okay, I think that's done.

awatson1978:
How are we doing on this?  Just ran into the error again today, and going through the a local patch instead of deploying this improved version of the library.

awatson1978:
2 weeks later, having to sloppy patch another server.  😞

Issue #1248, opened 2023-04-04
error TS2724: '"bson"' has no exported member named 'ObjectID'. Did you mean 'ObjectId'?


KeithGillette:
# Bug report

## Describe the bug
Running `tsc` from `typescript@4.9.5` with default `skipLibCheck: false` on a project importing ` @accounts/mongo@0.34.0` and `mongodb@5.1.0` generates the following error:
```
error TS2724: '"bson"' has no exported member named 'ObjectID'. Did you mean 'ObjectId'?
```

## To Reproduce
See bug description.

## Expected behavior
The project should transpile without error with TypeScript's default `skipLibCheck: false` setting.

## Screenshots
N/A

## System information
* accounts-js: 0.33.2
* mongodb: 5.1.0
* typescript: 4.9.5

## Additional context
The version of `mongodb` in `@accounts/mongo` dependencies is 2 major versions behind.

darkbasic:
Please try [1.0](https://github.com/accounts-js/accounts/pull/1258), this should be fixed.

Issue #1247, opened 2022-12-22
jsonwebtoken vulnerability


acomito:
![Screen Shot 2022-12-22 at 10 26 14 AM](https://user-images.githubusercontent.com/12701635/209167174-45d8dd50-1c2e-4209-8fcf-de73f987d704.png)

Issue #1244, opened 2022-11-26
Add graphql 16.6.0 to peerDependencies in @accounts/graphql-api


danyhiol:
# Bug report

## Describe the bug

Appolo recently released `@apollo/server v4`. Doing this, they dropped support for `Graphql < 16.6.0`.
`accounts-graphql-api` uses as peerDependencies `"graphql": "^14.6.0 || ^15.0.0",` this causes all kinds of problems working with the new apollo-server version.

## To Reproduce

Try installing @apollo/server@4.0.0 when using `@accounts/graphql-api`

## Expected behavior

Avoid dependencies conflict by updating graphql.

## Screenshots
`> npm ls graphql`
 ```js
@accounts/graphql-api@0.33.2
│ ├─┬ @graphql-tools/merge@6.2.13
│ │ ├─┬ @graphql-tools/schema@7.1.5
│ │ │ └── graphql@15.8.0 deduped invalid: "^16.6.0" from node_modules/@apollo/server, "^16.6.0" from node_modules/type-graphql
│ │ └── graphql@15.8.0 deduped invalid: "^16.6.0" from node_modules/@apollo/server, "^16.6.0" from node_modules/type-graphql
│ ├─┬ @graphql-tools/utils@7.9.0
│ │ └── graphql@15.8.0 deduped invalid: "^16.6.0" from node_modules/@apollo/server, "^16.6.0" from node_modules/type-graphql
│ └── graphql@15.8.0 deduped
├─┬ @accounts/graphql-client@0.33.1
│ ├─┬ @graphql-typed-document-node/core@3.1.0
│ │ └── graphql@15.8.0 deduped invalid: "^16.6.0" from node_modules/@apollo/server, "^16.6.0" from node_modules/type-graphql
│ └── graphql@15.8.0 deduped
├─┬ @apollo/client@3.6.9
│ ├─┬ @graphql-typed-document-node/core@3.1.1
│ │ └── graphql@15.8.0 deduped invalid: "^16.6.0" from node_modules/@apollo/server, "^16.6.0" from node_modules/type-graphql
│ └── graphql@15.8.0 deduped
├─┬ @apollo/server@4.2.1
│ ├─┬ @apollo/cache-control-types@1.0.2
│ │ └── graphql@15.8.0 deduped invalid: "^16.6.0" from node_modules/@apollo/server, "^16.6.0" from node_modules/type-graphql
│ ├─┬ @apollo/server-gateway-interface@1.0.7
[...]

Issue #1226, opened 2022-06-27
error TS2430: Interface 'ExecutionResult<TData>' incorrectly extends interface 'ExecutionResult<ObjMap<unknown>, ObjMap<unknown>>'.


KeithGillette:
# Bug report

## Describe the bug
Running `tsc` from `typescript@4.7.4` with default `skipLibCheck: false` on a project importing `@graphql/api@0.33.2` and `graphql@16.5.0` generates the following error:

```
node_modules/@accounts/graphql-api/node_modules/@graphql-tools/utils/Interfaces.d.ts:3:18 - error TS2430: Interface 'ExecutionResult<TData>' incorrectly extends interface 'ExecutionResult<ObjMap<unknown>, ObjMap<unknown>>'.
  Types of property 'data' are incompatible.
    Type 'TData' is not assignable to type 'ObjMap<unknown>'.

3 export interface ExecutionResult<TData = Record<string, any>> extends GraphQLExecutionResult {
```
## To Reproduce
See bug description.

## Expected behavior
The project should transpile without error with TypeScript's default `skipLibCheck: false` setting.

## Screenshots
N/A

## System information

- accounts-js: 0.33.2
- graphql: 16.5.0
- typescript: 4.7.4

## Additional context
The version of `graphql` in `@accounts/graphql-api` dependencies is 2 major versions behind.

KeithGillette:
The outdated `graphql` peer dependency also results in `npm install` errors with `npm>=8.6`when attempting to use the latest `graphql`:

```
npm ERR! Could not resolve dependency:
npm ERR! peer graphql@"^14.6.0 || ^15.0.0" from @accounts/graphql-api@0.33.2
npm ERR! node_modules/@accounts/graphql-api
npm ERR!   @accounts/graphql-api@"0.33.2" from the root project
```

PR #1258, opened 2023-10-22
feat: release 1.0

darkbasic:
Closes https://github.com/accounts-js/accounts/issues/1248
Closes https://github.com/accounts-js/accounts/issues/1247
Closes https://github.com/accounts-js/accounts/issues/1244
Closes https://github.com/accounts-js/accounts/issues/1226
Closes https://github.com/accounts-js/accounts/issues/1156
Closes https://github.com/accounts-js/accounts/issues/1142
Closes https://github.com/accounts-js/accounts/issues/1034
Closes https://github.com/accounts-js/accounts/issues/816
Closes https://github.com/accounts-js/accounts/issues/768
Closes https://github.com/accounts-js/accounts/issues/632
Closes https://github.com/accounts-js/accounts/issues/627
Closes https://github.com/accounts-js/accounts/issues/607
Closes https://github.com/accounts-js/accounts/issues/483
Closes https://github.com/accounts-js/accounts/issues/466
Closes https://github.com/accounts-js/accounts/issues/404

### DIFF
--- a/.changeset/afraid-rules-heal.md
+++ b/.changeset/afraid-rules-heal.md
@@ -1,0 +1,10 @@
+---
+'@accounts/mikro-orm': patch
+'@accounts/typeorm': patch
+'@accounts/express-session': patch
+'@accounts/magic-link': patch
+'@accounts/password': patch
+'@accounts/server': patch
+---
+
+Upgrade lodash to 4.17
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -7,6 +7,7 @@
   "baseBranch": "master",
   "updateInternalDependencies": "patch",
   "ignore": [
+    "accounts-js",
     "@accounts/e2e",
     "@examples/accounts-microservice",
     "@examples/graphql-mikro-orm-typescript",
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -8,13 +8,6 @@
   "updateInternalDependencies": "patch",
   "ignore": [
     "@accounts/e2e",
-    "@accounts/mikro-orm",
-    "@accounts/module-core",
-    "@accounts/module-magic-link",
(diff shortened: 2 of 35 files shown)


mrcleanandfresh:
Okay, looks like your example needed:

```javascript
import 'reflect-metadata';
```

At the top. I'll put that on my list.

darkbasic:
Yes you need reflect-metadata. If something from the documentation doesn't work I suggest you to have a look at the examples folder because these get tested much more often.
Docs need to be updated to reflect some of the latest changes.

mrcleanandfresh:
> Yes you need reflect-metadata. If something from the documentation doesn't work I suggest you to have a look at the examples folder because these get tested much more often. Docs need to be updated to reflect some of the latest changes.

Yeah, my plan is to take the documentation for a test drive and see if there are any holes in it for folks getting up and running. I was able to get up-and-running in about 30-45 minutes after some troubleshooting. So I took note of those missing elements. Most folks are going to want as little friction as possible while on-boarding. So I am keeping a list of changes I want to make to the documentation.

Good to know about the examples folder! Thanks.

Maybe we should leave comments in the open issues about the new release. I saw a very serious bug with a failed OAuth attempt that returned the first record from the users database. That's a pretty serious bug, but may be a non-issue after the re-factor? It'd be good to have the issue authors go back through after the official release to see if they are still experiencing their problems. Thoughts?

darkbasic:
> Maybe we should leave comments in the open issues about the new release. I saw a very serious bug with a failed OAuth attempt that returned the first record from the users database. That's a pretty serious bug, but may be a non-issue after the re-factor?

Yes leave comments in the respective issues if you find anything that you can still reproduce.
I saw the OAuth one and I honestly didn't check it yet, but that's surely one that needs to be addressed before the stable release.

